PR #7, opened 2018-05-29
Remove useless AppBarLayout in GardenActivity layout

Syex:
The layout file for this activity `activity_garden.xml` contained two `AppBarLayouts`, but one was never used and also useless for the layout.


tiembo:
Thanks for the pull request!  We've addressed this internally and have pushed the changes to this repo.

PR #18, opened 2018-06-04
Use Fragment.instantiate instead new-call on fragment

XinyueZ:


tiembo:
Thanks for the PR. Let's keep the code as-is to follow the example outlined in the [Fragments guide](https://developer.android.com/guide/components/fragments#Example).

XinyueZ:
if dont use `instantiate` , it could lost `bundle` if the `fragment` was killed by runtime.

PR #19, opened 2018-06-04
Used extension style of kotlin for binding-adapter

XinyueZ:
It brings a good manner that people'd use these methods not only for
    binding, but also for normal use, better than java-style static
    methods.


tiembo:
I do like this approach, but have a slight preference in keeping the methods similar to the [binding adapter documentation](https://developer.android.com/topic/libraries/data-binding/binding-adapters).

Issue #9, opened 2018-05-29
Why having multiple activities?


gderaco:
According to [this](https://android-developers.googleblog.com/2018/05/use-android-jetpack-to-accelerate-your.html) article, Google suggested using only one Activity: 

_Today we are introducing the Navigation component as a framework for structuring your in-app UI, with a focus on making a single-Activity app the preferred architecture. With out of the box support for Fragments, you get all of the Architecture Components benefits such as Lifecycle and ViewModel while allowing Navigation to handle the complexity of FragmentTransactions for you._ 

So my question is, why didn't you follow the advice yourself gave?

ShiroYacha:
It does look a bit counter-intuitive that the "master-detail" view is split into 2 activities whereas the less related views are 2 fragments of the same view...

XinyueZ:
why not :)

tiembo:
That's correct - a single-Activity app is preferred.  We'll be consolidating down to one activity.

XinyueZ:
@tiembo @gderaco   Navigation within "single-activity" is really a good way to control application route and move direction, do it also, however, my point of view is that activity should capsulate "business"  group. A typical example is that user-shop-buying. 

I will user-profile with edit, list, view in one activity.
For shop, list, detail and gallery of item if needed in  one activity.
Buying process in one activity.

Flexible with multi-single-activity is my variant of jetpack. This is my opinion

PR #21, opened 2018-06-04
Optimized code for opening PlantDetailActivity

XinyueZ:
Like newInstance of fragments, here I introduce same pattern on opening
    the activities. For code-reader it helps us to find target quickly.


tiembo:
Per #9, the project will be consolidated down to one Activity.

PR #22, opened 2018-06-04
Add CircleCI

tiembo:


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+    environment:
+      JVM_OPTS: -Xmx3200m
+      TERM: dumb
+    steps:
+      - checkout
+      - restore_cache:
+          <<: *cache_key
+      - run:
+          name: Download Dependencies
+          command: ./gradlew androidDependencies
```
chrisbanes: I'd just use `./gradlew dependencies`. This will pull in all plugins and classpath deps too.

.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+      - save_cache:
+          <<: *cache_key
+          paths:
+            - ~/.gradle
```
chrisbanes: Pulling in all of `~/.gradle` is overkill (and can get large). That's why KTX + Tivi only save certain folders.

.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+      - run:
+          name: Run JVM Tests & Lint
+          command: ./gradlew check
+      - store_artifacts:
+          path: app/build/reports
```
chrisbanes: Do you not want to store the APKs?

.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+references:
+  cache_key: &cache_key
+    key: jars-{{ checksum "build.gradle" }}-{{ checksum  "app/build.gradle" }}-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+
+jobs:
+  build:
+    docker:
+      - image: circleci/android:api-27-alpha
+    environment:
+      JVM_OPTS: -Xmx3200m
```
chrisbanes: Way too much memory allocated here. Circle has a 4G limit, but here you're giving each JVM 3.2GB. Gradle by default will create workers = CPU count, so that's 12GB. When R8 becomes default, that will be done out of process too.

I'd suggest limiting your workers and max heap. See my [Tivi](https://github.com/chrisbanes/tivi/blob/master/.circleci/config.yml#L8) for an example. Also disable the daemon as that uses RAM and is unnecessary for CI.

.circleci/config.yml
tiembo: Not at this point - just build and test for now.

.circleci/config.yml
tiembo: Done.

.circleci/config.yml
tiembo: Done - only cache `~/.gradle/caches` and `~/.gradle/wrapper`.

.circleci/config.yml
tiembo: Good to know.  Tweaked environment settings.

PR #26, opened 2018-06-05
Fixed static-import of view to use GONE

XinyueZ:


tiembo:
Is this PR meant to just fix the static import of `GONE`? There are a lot of other changes attached.

XinyueZ:
@tiembo I think this branch is not needed more
I closed it.

Issue #20, opened 2018-06-04
Why not data-binding in GardenPlantingAdapter?


XinyueZ:
I see still there are some `findViewById` calls and use `non-data-binding` style.
Is that the contract of this project that `data-binding` would not be used in list-view like objects?
I want a confirm, otherwise I will try a PR to optimize it with `data-binding`.

tiembo:
Yes, the adapters should also use data binding.  That'd be great if you could open a PR for data binding in `GardenPlantingAdapter`.

PR #28, opened 2018-06-05
Refactored PlantAdapter with data-binding

XinyueZ:
According to ticket #20 as well.


app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
```diff
@@ -30,7 +30,7 @@ import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
 import com.google.samples.apps.sunflower.R
 import com.google.samples.apps.sunflower.data.PlantAndGardenPlantings
 import java.text.SimpleDateFormat
-import java.util.Locale
+import java.util.*
```
tiembo: nit: no wildcard imports

XinyueZ: @tiembo  solved

PR #27, opened 2018-06-05
Refactored GardenPlantingAdapter with data-binding

XinyueZ:
According to ticket #20 .

PR #45, opened 2018-06-14
Slight refactor to Plant adapter and data binding

tiembo:
Set data binding variables in a `bind` method in the ViewHolder, following the guidance from the [Android Data Binding: RecyclerView article](https://medium.com/google-developers/android-data-binding-recyclerview-db7c40d9f0e4).

cc @XinyueZ

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
@@ -17,7 +17,6 @@
 package com.google.samples.apps.sunflower.adapters
 
 import android.content.Intent
-import android.databinding.ViewDataBinding
 import android.support.v7.recyclerview.extensions.ListAdapter
 import android.support.v7.widget.RecyclerView
 import android.view.LayoutInflater
@@ -56,11 +55,14 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
                 LayoutInflater.from(parent.context), parent, false))
     }
 
-    class ViewHolder(private val binding: ViewDataBinding) : RecyclerView.ViewHolder(binding.root) {
-        fun bind(clickListener: View.OnClickListener, item: Plant) {
+    class ViewHolder(
+            private val binding: ListItemPlantBinding
+    ) : RecyclerView.ViewHolder(binding.root) {
+
+        fun bind(listener: View.OnClickListener, item: Plant) {
             binding.apply {
-                setVariable(BR.clickListener, clickListener)
-                setVariable(BR.plant, item)
+                clickListener = listener
+                plant = item
                 executePendingBindings()
             }
         }
--- a/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
@@ -16,10 +16,7 @@
 
 package com.google.samples.apps.sunflower.adapters
 
-import android.arch.lifecycle.ViewModel
 import android.content.Intent
-import android.databinding.DataBindingUtil
-import android.databinding.ObservableField
 import android.databinding.ViewDataBinding
 import android.support.v7.recyclerview.extensions.ListAdapter
 import android.support.v7.widget.RecyclerView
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
```diff
@@ -47,29 +44,23 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
...
-            holder.itemView.tag = this
-            with(holder.binding) {
-                setVariable(BR.vm, ItemViewModel(this@apply, onClickListener))
-                executePendingBindings()
-            }
+        val plant = getItem(position)
+        holder.apply {
+            bind(plant)
+            itemView.tag = plant
+            itemView.setOnClickListener(onClickListener)
```
XinyueZ: with `bind()` is ok, however, I can't understand why keep `listener` out of binding instead in xml?

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
```diff
@@ -47,29 +44,23 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
...
-        val content = ObservableField<String>(plant.name)
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+        return ViewHolder(ListItemPlantBinding.inflate(
+                LayoutInflater.from(parent.context), parent, false))
+    }
 
-        val imageUrl = ObservableField<String>(plant.imageUrl)
+    class ViewHolder(private val binding: ViewDataBinding) : RecyclerView.ViewHolder(binding.root) {
+        fun bind(item: Plant) {
+            binding.setVariable(BR.plant, item)
```
XinyueZ:
```
with(binding){
...
}
```

app/src/main/res/layout/list_item_plant.xml
```diff
@@ -20,16 +20,14 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <data>
-
         <variable
-            name="vm"
-            type="com.google.samples.apps.sunflower.adapters.PlantAdapter.ItemViewModel"/>
+            name="plant"
+            type="com.google.samples.apps.sunflower.data.Plant"/>
```
XinyueZ: I'm not so strictly against direct connection between domain and UI, however, I would only suggest using a view-model to a capsulate domain in order to isolate UI and logical with the view-model.

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
tiembo: Slight preference for `apply` since it's used above in this file.  For #27 we can use `with` throughout, which can illustrate the various options available in Kotlin.  What do you think?

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
tiembo: Reverted the change, so now the listener is passed as a binding variable.

app/src/main/res/layout/list_item_plant.xml
tiembo: In the 2018 version of the Google I/O Android App we used this pattern for RecyclerView list items.  I think a ViewModel is a bit heavyweight for this use case.

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
```diff
@@ -47,29 +44,25 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
...
-        ViewModel() {  // Actually no need, however, for unified namespace with other [ViewModels]s I consider it.
-
-        val content = ObservableField<String>(plant.name)
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+        return ViewHolder(ListItemPlantBinding.inflate(
+                LayoutInflater.from(parent.context), parent, false))
+    }
 
-        val imageUrl = ObservableField<String>(plant.imageUrl)
+    class ViewHolder(private val binding: ViewDataBinding) : RecyclerView.ViewHolder(binding.root) {
```
thagikura: How about writing like this?
It uses ListItemPlantBinding instead of ViewDataBinding so that variables can be accessed directly instead of using setVariable


```kt
    class ViewHolder(private val binding: ListItemPlantBinding) : RecyclerView.ViewHolder(binding.root) {
        fun bind(listener: View.OnClickListener, item: Plant) {
            binding.apply {
                clickListener = listener
                plant = item
                executePendingBindings()
            }
        }
    }
```

tiembo: Oh, neat! I applied your suggestion.  PTAL

Issue #47, opened 2018-06-14
Material Design specs


andhie:
The sample app UI is quite crude, and parts of it does not follow the MD specs.
e.g. the Add icon used in the FAB is using the wrong size
e.g. paddings of text

Are there any plans to touch up the UI? or a mock that i can follow for a PR

tiembo:
Yes, there are plans to add Material design.  Would you like to take a pass at it?

A good first PR for MD would be simply adding the dependency and changing the theme: https://material.io/develop/android/docs/getting-started/

andhie:
sure, i'll add in the dependency, Theme.MaterialComponent, and some minor xml UI tweaks

tiembo:
You might run into issues with implementing this.  Material Design Components require AndroidX, and there is currently an issue with AndroidX and Navigation.

andhie:
Temporarily we can use `com.android.support:design:28.0.0-alpha3` since its the same as `com.google.android.material:material`. 
but doing so will increase the `ext {supportLibraryVersion = '28.0.0-alpha3'}`. are you fine with this?

To fully migrate to androidx, i think it should be a commit on its own.

andhie:
There are a few things that require more touch up.

1) The spacing of the text between image and text in list item. they seems a little far, may be confusing due to the text might related to the top or the bottom image without reading the text.
suggestion 1: put the text on the image to inline it. text may need to be shorter to fit nicer
suggestion 2: have more space between the text and the next item (32dp), doing this will probably have to write a SpaceItemDecorator that applies top margin on 2nd item onwards
https://material.io/design/components/image-lists.html#anatomy

2) The image size should be in ratio. the recommended ratios are 16:9, 3:2, 4:3. the garden fragment items can be taller to support more text (e.g. 4:3 ?) while plant list can be 3:2.
https://material.io/design/layout/spacing-methods.html#containers-aspect-ratios

3) Doing transition animation to detail activity via the hero image

4) the text padding/margin in detail activity. 16dp instead of 8dp

ankuranurag2:
The navigation drawer should close first (if open) onBackPressed. Right now, the drawer is left open while fragment transactions going in the back.

tiembo:
@ankuranurag2 thanks for reporting that.  Can you create a new issue to track the bug separately from the this general MDC redesign issue?

andhie:
@tiembo i think we should close this thread and open individual issues instead of delaying v0.1.3.
[...]

PR #48, opened 2018-06-14
Applying Material Design

andhie:
1) Updated to support lib 28 to use new MD styles. 
2) padding/margin changes to fix overscroll visually being "cut" off
2.1) fix the list items having different visual spacing/paddings
3) Use Vector drawable of the plus icon. specs requires 24dp, but uses 32dp icon
4) Use the new updated MD 2018 specs via `Theme.MaterialComponents` and `Widget.MaterialComponents`. the default still uses `Widget.Design`
5) Translucent statusbar
5.1) support DrawerLayout above AppBar but below statusbar
5.2) support detail screen to have hero image below statusbar
6) disable drawer toggle animation as drawer opens
7) Animate FAB hide on add plant.

As a starting point to implement #47


app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantDetailBindingAdapter.kt
```diff
@@ -38,8 +38,8 @@ fun imageFromUrl(view: ImageView, imageUrl: String?) {
 }
 
 @BindingAdapter("goneIf")
-fun goneIf(view: View, isGone: Boolean) {
-    view.visibility = if (isGone) View.GONE else View.VISIBLE
+fun goneIf(view: FloatingActionButton, isGone: Boolean?) {
```
XinyueZ: I don't think ```goneIf``` should lose origin objective, because general ```views``` need it. 
On the contrary, I would like to add a new extension method for ```FloatingActionButton``` to hide.
How do you think? @andhie @tiembo

app/src/main/res/layout/list_item_plant.xml
```diff
@@ -51,14 +51,14 @@
...
-            android:layout_marginStart="@dimen/margin_normal"
+            android:layout_marginEnd="@dimen/margin_small"
+            android:layout_marginStart="@dimen/margin_small"
             android:layout_marginTop="@dimen/margin_small"
             android:text="@{vm.content}"
             android:textAppearance="?attr/textAppearanceListItem"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toBottomOf="@id/plant_item_image"
+            app:layout_constraintTop_toBottomOf="@+id/plant_item_image"
```
tiembo: +id isn't required here

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantDetailBindingAdapter.kt
tiembo: Originally this `goneIf` did serve the purpose of a general `GONE` / `VISIBLE` binding adapter.  But since `goneIf` is currently only used with the FAB (and is in `PlantDetailBindingAdapter`) I think it's appropriate for the param to be a FAB instead of a `View`.  This also allows the use of `hide()` and `show()`, giving us animations for free.

If there's a need for a general purpose `goneIf` we can place it in a generic `BindingAdapter` file.

app/src/main/res/layout/list_item_plant.xml
andhie: i know it doesnt, but adding it 'fixes' the preview designer to show it below the image view

app/src/main/java/com/google/samples/apps/sunflower/GardenActivity.kt
```diff
@@ -50,6 +50,7 @@ class GardenActivity : AppCompatActivity() {
         drawerLayout = findViewById(R.id.drawer_layout)
         drawerToggle = ActionBarDrawerToggle(
                 this, drawerLayout, R.string.drawer_open, R.string.drawer_close)
+        drawerToggle.isDrawerSlideAnimationEnabled = false
```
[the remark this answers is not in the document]
andhie: there are no reason to enable it. 

1) the specs did not mention about animation
2) it would have hidden by the drawer anyway
3) historically, it was used for a short time until the official support lib v21 released. which suggest the use of `Toolbar` with `setNavigationIcon` and `setNavigationOnClickListener` which is more simplified than using drawer toggle

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantDetailBindingAdapter.kt
andhie: at first i actually duplicated the exact same method name and bindingadapter. although Data Binding smartly knows which method to pick (`View` vs `FloatingActionButton` param), this causes the `View` goneif method to grey out as it was not used. plus, this is a sample app, i wish to keep it simple enough and less confusing to newbies

app/src/main/res/layout/list_item_plant.xml
tiembo: I'm unable to reproduce this with Android Studio 3.2 Canary 18.  Try Invalidating Caches / Restart from the Android Studio File menu and see if that fixes things.

Adding a + also introduces a bug by adding itself as an option for "Go to declaration".  I think we should remove the +.

andhie: added another commit to remove it

tiembo:
Thanks @andhie !

PR #50, opened 2018-06-15
Use Fragment viewLifecycleOwner

### LINKED ISSUE #36: Bug : Multiple subscription to livedata
Every time you switch fragment from GardenPlantListing to PlantListing and vice versa both fragments are detached and re-attached causing **onCreateView** call. Since in onCreateView we call subscribeUi method causing **multiple live data observer**. This will lead to multiple callbacks for livedata observer and because of this multiple **notifyDataSetChanged** in adapter.

This blog post explain the issue with livedata and fragment

https://medium.com/@BladeCoder/architecture-components-pitfalls-part-1-9300dd969808

To avoid multiple observer we need to use **new view life cycle** while observing livedata instead of fragment lifecycle.

Following session from google I/O 18 will help to understand the issue better

https://youtu.be/pErTyQpA390?t=5m41s

New view life cycle is added in support lib 28.0.0 and androidx 1.0.0.

andhie:
Fix #36

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
@@ -48,7 +48,7 @@ class GardenFragment : Fragment() {
         val viewModel = ViewModelProviders.of(this, factory)
                 .get(GardenPlantingListViewModel::class.java)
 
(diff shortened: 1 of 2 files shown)


tiembo:
Thanks

PR #51, opened 2018-06-15
Update versions of kotlin and constrain-layout

XinyueZ:


build.gradle
```diff
@@ -32,7 +32,7 @@ buildscript {
         junitVersion = '4.12'
         lifecycleVersion = '1.1.1'
         navigationVersion = '1.0.0-alpha02'
-        kotlinVersion = '1.2.41'
+        kotlinVersion = '1.2.50'
```
tiembo: The Kotlin update is addressed in #52.  Can you remove this change?

XinyueZ:
@tiembo plz check out

PR #52, opened 2018-06-15
Update Kotlin to 1.2.50 (with workarounds)

tiembo:
This change includes a temporary workaround for a Kotlin 1.2.50 and DataBinding bug.

For more context see https://issuetracker.google.com/issues/110198434 and https://youtrack.jetbrains.com/issue/KT-24915#comment=27-2914947

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,9 +33,29 @@ android {
         versionName "0.1.2"
     }
     buildTypes {
+        debug {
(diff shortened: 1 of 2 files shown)


XinyueZ:
@tiembo The workaround has a problem when the application builds with multi-flavors.

tiembo:
@XinyueZ can you open up a new issue? We can continue discussion there.

PR #57, opened 2018-06-20
Fix TODO (auto-generate ID) in  GardenPlanting(including use and tests)

XinyueZ:


XinyueZ:
@tiembo I use `long` instead `string` to use SQLite build-In feature to generate automatic ids.

app/src/test/java/com/google/samples/apps/sunflower/data/GardenPlantingTest.kt
```diff
@@ -26,8 +26,9 @@ import java.util.Calendar.YEAR
 
 class GardenPlantingTest {
 
-    @Test fun testDefaultValues() {
-        val gardenPlanting = GardenPlanting("1", "1")
+    @Test
+    fun testDefaultValues() {
+        val gardenPlanting = GardenPlanting("1").also { it.gardenPlantingId = 1 }
```
tiembo: Remove the `.also` block and add a test: `assertEquals(0L, gardenPlanting.gardenPlantingId)`

app/src/androidTest/java/com/google/samples/apps/sunflower/data/GardenPlantingDaoTest.kt
```diff
@@ -48,14 +48,18 @@ class GardenPlantingDaoTest {
...
+            testCalendar
+        ).also { it.gardenPlantingId = 2 }
         gardenPlantingDao.insertGardenPlanting(gardenPlanting2)
         assertThat(getValue(gardenPlantingDao.getGardenPlantings()).size, equalTo(2))
     }
 
     @Test fun testGetGardenPlanting() {
         assertThat(getValue(gardenPlantingDao.getGardenPlanting(
-                testGardenPlanting.gardenPlantingId)), equalTo(testGardenPlanting))
+                testGardenPlanting.gardenPlantingId.toString())), equalTo(testGardenPlanting))
```
tiembo: Remove the `.toString()` and change `GardenPlantingDao#getGardenPlanting` to take in a `Long` instead of a `String`.

app/src/test/java/com/google/samples/apps/sunflower/data/GardenPlantingTest.kt
XinyueZ: Yes, you're right, the id should have to be given automatically.

app/src/androidTest/java/com/google/samples/apps/sunflower/data/GardenPlantingDaoTest.kt
XinyueZ: Updated

XinyueZ:
@tiembo In order to get "auto-generated id" after inserting new data I've given return `Long` to ` insertGardenPlanting(gardenPlanting: GardenPlanting)` , also I have updated tests for this change.

tiembo:
Thank you!

PR #74, opened 2018-06-29
Let dataBindingVersion equal to gradleVersion

XinyueZ:


tiembo:
The `dataBindingVersion` variable actually isn't used.  Let's remove it instead.

XinyueZ:
@tiembo you're right, I 've deleted it, the binding should always default.

tiembo:
Thanks!

PR #78, opened 2018-07-02
Fix #75:  an approach to handle duplicated "appbar"

### LINKED ISSUE #75: Detail page has two toolbars
After converting to single-activity, there're two `toolbar`s on detail screen.
The app needs a ux/ui decision what should be shown in detail page. @tiembo

XinyueZ:
- This PR would fix the bug that the detail page shows two appbars which one of them come from main and another from detail itself. 
- The approach is still using appbar of detail but using a "back" icon for navi-up of navi-controller. 
- A fade-in/out effect has been used while opening item from main to detail.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/Extensions.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/Extensions.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2018 Google LLC
+ *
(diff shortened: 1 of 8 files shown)


tiembo:
Closed - see [comment](https://github.com/googlesamples/android-sunflower/issues/75#issuecomment-459554681) for more details.

PR #81, opened 2018-07-02
Add Kotlin lint via ktlint

tiembo:
This PR also:
* Adds instructions for configuring ktlint for Android Studio
* Applies ktlint fixes using `./gradlew sA`


tiembo:
@thagikura Thanks!

The `sA` task is short for `spotlessApply`, which fixes the style violations.

PR #84, opened 2018-07-03
Fix #75: alternative approach to handle duplicated "appbar"

### LINKED ISSUE #75: Detail page has two toolbars

XinyueZ:
This is the alternative solution (against #78) for case #75.  @tiembo.  The approach is using different `toolbars` for different `fragments`. Because some general codes should not be used duplicate, there's a new file` Extensions.kt `which brings `setUpAppbar() `for those `toolbars`. 

Some layouts will be significantly altered, because of toolbars.

Toolbar feature-related IDs are defined in` Ids.xml.` Because `GradenFragment` as home, must know the existence of drawer_layout, so I chose, everything that navigation has associated is defined generically.


app/src/main/java/com/google/samples/apps/sunflower/utilities/Extensions.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.support.v7.app.AppCompatActivity
+import android.view.View
+import androidx.navigation.Navigation
+import androidx.navigation.ui.NavigationUI
+import com.google.samples.apps.sunflower.R
+
+fun AppCompatActivity.setUpAppbar(view: View, isHome: Boolean) {
+    setSupportActionBar(view.findViewById(R.id.toolbar))
+    if (isHome) {
+        NavigationUI.setupActionBarWithNavController(
```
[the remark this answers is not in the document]
XinyueZ: @sagar-viradiya I have pushed a commit(87020d7) which set up `appbar` with normal approach without extension. However, I want to know some other opinions on this change, because I really mean to save duplicated codes with extension approach(5e46365).

app/src/main/java/com/google/samples/apps/sunflower/utilities/Extensions.kt
```diff
@@ -1,39 +0,0 @@
...
-import android.support.v7.app.AppCompatActivity
-import android.view.View
-import androidx.navigation.Navigation
-import androidx.navigation.ui.NavigationUI
-import com.google.samples.apps.sunflower.R
-
-fun AppCompatActivity.setUpAppbar(view: View, isHome: Boolean) {
-    setSupportActionBar(view.findViewById(R.id.toolbar))
-    if (isHome) {
-        NavigationUI.setupActionBarWithNavController(
```
sagar-viradiya: I am fine with extension. Just replace setupActionBarWithNavController with normal way to set action bar. I didn't mean to move logic to individual fragment.

XinyueZ: @sagar-viradiya sry, can't catch ur meaning. Any dummy?

tiembo:
Closed - see [comment](https://github.com/googlesamples/android-sunflower/issues/75#issuecomment-459554681) for more details.

PR #86, opened 2018-07-04
Share detail

### LINKED ISSUE #82: Add Sharing to the plant detail page
Incorporate the [Sharing Jetpack component](https://developer.android.com/training/sharing/shareaction) to the plant detail page.

andhie:
Fix #82

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
@@ -50,10 +50,11 @@ class PlantDetailFragment : Fragment() {
 
         val factory = InjectorUtils.providePlantDetailViewModelFactory(requireActivity(), plantId)
         plantDetailViewModel = ViewModelProviders.of(this, factory)
-                .get(PlantDetailViewModel::class.java)
+            .get(PlantDetailViewModel::class.java)
 
         val binding = DataBindingUtil.inflate<FragmentPlantDetailBinding>(
-                inflater, R.layout.fragment_plant_detail, container, false).apply {
+            inflater, R.layout.fragment_plant_detail, container, false
+        ).apply {
             viewModel = plantDetailViewModel
             setLifecycleOwner(this@PlantDetailFragment)
             fab.setOnClickListener { view ->
@@ -75,20 +76,19 @@ class PlantDetailFragment : Fragment() {
     override fun onOptionsItemSelected(item: MenuItem?): Boolean {
         return when (item?.itemId) {
             R.id.action_share -> {
-                share()
+                val shareText = getString(R.string.share_text, plantDetailViewModel.plant.value?.name)
+                startActivity(getShareTextIntent(shareText))
                 return true
             }
             else -> super.onOptionsItemSelected(item)
         }
     }
 
-    private fun share() {
-        val shareText = getString(R.string.share_text, plantDetailViewModel.plant.value?.name)
-
+    private fun getShareTextIntent(text: String): Intent {
         val intent = ShareCompat.IntentBuilder.from(activity)
-                .setText(shareText)
-                .setType("text/plain")
-                .createChooserIntent()
+            .setText(text)
+            .setType("text/plain")
+            .createChooserIntent()
 
         // https://android-developers.googleblog.com/2012/02/share-with-intents.html
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
@@ -99,7 +99,7 @@ class PlantDetailFragment : Fragment() {
             intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET)
         }
 
-        startActivity(intent)
+        return intent
     }
 
     companion object {
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
@@ -17,14 +17,19 @@
 package com.google.samples.apps.sunflower
 
 import android.arch.lifecycle.ViewModelProviders
+import android.content.Intent
 import android.databinding.DataBindingUtil
+import android.os.Build
 import android.os.Bundle
 import android.support.design.widget.Snackbar
 import android.support.v4.app.Fragment
+import android.support.v4.app.ShareCompat
 import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
-
 import com.google.samples.apps.sunflower.databinding.FragmentPlantDetailBinding
 import com.google.samples.apps.sunflower.utilities.InjectorUtils
 import com.google.samples.apps.sunflower.viewmodels.PlantDetailViewModel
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -53,11 +60,48 @@ class PlantDetailFragment : Fragment() {
...
+        return when (item?.itemId) {
+            R.id.action_share -> {
+                share()
+                return true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun share() {
```
XinyueZ: It could be a wonder that the "share"  will be packed as an `extension` of `activity` or `fragment` which has title and content as arguments. As learning project for people who want to get some reusable codes, the extension can help more. How do you think?

XinyueZ:
@andhie Nice2have to bring some TESTs to this new feature what I want to suggest.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
andhie: i dont think "share" should tie to an `Activity` or `Fragment`. this is similar to String case where people wanted to extend it. but would it be appropriate? 
i'm not too sure how you do reusable code in kotlin, in java there is always a "Util" class

andhie: lemme change it to `fun getShareTextIntent(text: String): Intent` for more flexible use case.

andhie:
added test

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -42,22 +49,59 @@ class PlantDetailFragment : Fragment() {
...
+            inflater, R.layout.fragment_plant_detail, container, false
+        ).apply {
             viewModel = plantDetailViewModel
             setLifecycleOwner(this@PlantDetailFragment)
             fab.setOnClickListener { view ->
                 plantDetailViewModel.addPlantToGarden()
                 Snackbar.make(view, R.string.added_plant_to_garden, Snackbar.LENGTH_LONG).show()
             }
+
+            setHasOptionsMenu(true)
```
tiembo: Should be outside of `.apply`

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -42,22 +49,59 @@ class PlantDetailFragment : Fragment() {
...
+    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {
+        inflater?.inflate(R.menu.menu_plant_detail, menu)
+        super.onCreateOptionsMenu(menu, inflater)
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem?): Boolean {
+        return when (item?.itemId) {
+            R.id.action_share -> {
+                val shareText = getString(R.string.share_text, plantDetailViewModel.plant.value?.name)
+                startActivity(getShareTextIntent(shareText))
```
tiembo: No need for a separate method.  This line can simply be:

```
ShareCompat.IntentBuilder.from(activity)
    .setText(shareText)
    .setType("text/plain")
    .startChooser()
```

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -42,22 +49,59 @@ class PlantDetailFragment : Fragment() {
...
+                return true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun getShareTextIntent(text: String): Intent {
```
tiembo: remove method

app/src/main/res/menu/menu_plant_detail.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tiembo: This line should be below the copyright.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
andhie: the separate method is for me to add the flags without making the method too long

tiembo:
@andhie here's a [gist of `PlantDetailFragment`](https://gist.github.com/tiembo/67e72bccf539fcf621370e397f7f15a3) constructing the share string using `.observe` instead of `.value`.  Please take a look!

app/src/main/res/menu/menu_plant_detail.xml
andhie: 1) i copied from existing menu.xml in the project and edited the item info only.
2) if you move this line below copyright, AS will have an error 'Xml declaration should precede all document content'. you wont be able to compile it

tiembo: My mistake - you're correct.  No action needed here.

andhie:
i've updated the code to use observer to build the string.
but the share method can still be debated. currently there is no way to addFlags in the builder, so to add flags, i have to build() the intent first. and the flags added are different for <21 and 21+. 
my opinion on separate method is aesthetics and easier to read the `onOptionsItemSelected` method

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
XinyueZ: These can save lines:
```
  private fun getShareTextIntent(text: String) = ShareCompat.IntentBuilder.from(activity)
        .setText(text)
        .setType("text/plain")
        .createChooserIntent().apply {
            // https://android-developers.googleblog.com/2012/02/share-with-intents.html
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
                // If we're on Lollipop, we can open the intent as a document
                addFlags(Intent.FLAG_ACTIVITY_NEW_DOCUMENT or Intent.FLAG_ACTIVITY_MULTIPLE_TASK)
            } else {
                // Else, we will use the old CLEAR_WHEN_TASK_RESET flag
                addFlags(Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET)
            }
        }
```

andhie: i like the apply function

tiembo:
Can you elaborate on the use of `FLAG_ACTIVITY_MULTIPLE_TASK`?

Also, `FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET` [performs exactly the same](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET) as `FLAG_ACTIVITY_NEW_DOCUMENT`, so the new flag can be used instead.

andhie:
First you would have to understand why we need the flags in the 1st place. Dianne Hackborn's post explains in detail https://plus.google.com/+DianneHackborn/posts/4QWEQgkB1v2.
i think this line sums up my intention 

> the meaning of the flag is, "hey system, yeah I know I am launching this thing as part of my task, but to the user it isn't really part of my app, so if they re-launch me from home please get rid of it or the user is doing to be very confused."

These flags are also implemented in `ShareActionProvider`
https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/v7/appcompat/src/main/java/android/support/v7/widget/ShareActionProvider.java#409

FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET performs exactly the same as FLAG_ACTIVITY_NEW_DOCUMENT
That is true, but are you forgetting the app is `minSdkVersion=19` ?

tiembo:
Thank you very much for the excellent writeup @andhie! I'll open a ticket internally to explore enhancing our share documentation with Dianne's post.

PR #89, opened 2018-07-05
Refactor: code formatting

williamwue:
Use "@id" may be more clear about FAB "layout_anchor" for existing view element. 
If someone remove "plant_detail_scrollview" view element and not change FAB, then compiler would report error to pointer out the reason for developer.


williamwue:
I signed it!

tiembo:
Thank you @williamwue!

PR #97, opened 2018-07-10
Fix #96: A potential mem-leak when call providePlantDetailViewModelFactory

### LINKED ISSUE #96: Will passing context to InjectorUtils singleton object cause memory leaks
If we send activity/fragment context to InjectorUtils singleton object, will they hold a reference to it even after they've been destroyed and lead to memory leaks?

Please clarify

XinyueZ:
See #96


andhie:
wouldnt it be better to enforce passing in `Application`? by changing all the InjectorUtils params to accept `Application` instead of just `Context`
we can see this style in `AndroidViewModel` that enforces to pass `Application`
https://developer.android.com/reference/androidx/lifecycle/AndroidViewModel#androidviewmodel

yshrsmz:
why not just call `context.applicationContext` inside `InjectorUtils`?

XinyueZ:
@andhie I want to here code against the interface(Context) instead of a concrete class. thank u for the tip @yshrsmz

XinyueZ:
I have resolved conflicts .

nic0lette:
Thanks again for the contribution. :)

PR #107, opened 2018-07-16
Code optimise, using expression instead block body

XinyueZ:
- Changes on `onCreateViewHold` of `GradenPlantingAdapter` and `PlantAdapter`


tiembo:
Thanks for the PR.  Per the [style guide](https://android.github.io/kotlin-guides/style.html#expression-functions) - since the expression functions wrap, they should be in a normal function block.

PR #108, opened 2018-07-16
Fix Room FK warnings

andhie:
> w:app/build/tmp/kapt3/stubs/debug/com/google/samples/apps/sunflower/data/GardenPlanting.java:15: warning: plant_id column references a foreign key but it is not part of an index. This may trigger full table scans whenever parent table is modified so you are highly advised to create an index that covers this column.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/data/AppDatabase.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/data/AppDatabase.kt
@@ -21,7 +21,6 @@ import android.arch.persistence.room.Database
 import android.arch.persistence.room.Room
 import android.arch.persistence.room.RoomDatabase
 import android.arch.persistence.room.TypeConverters
-import android.arch.persistence.room.migration.Migration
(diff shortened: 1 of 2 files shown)


tiembo:
Thanks for the PR!  Since Sunflower is currently in alpha let's omit the migration and keep just the FK. 
 The DB and schema will likely change often so migrations should only be used once the project is stable.

I'll update the README to explicitly point out non-backwards compatible changes while the project is in alpha.

andhie:
would you prefer to use `fallbackToDestructiveMigration` instead?

tiembo:
While adding `fallbackToDestructiveMigration` would work and prevent the app from crashing, I'm concerned that it'll be forgotten and overlooked, causing some odd behavior in the future.

I added the warning about not maintaining backwards compatibility in #118.  If it's OK with you let's merge this PR with just the FK fix.

andhie:
no problem for me. do merge

tiembo:
Thanks!

PR #109, opened 2018-07-16
Fix position string warning

andhie:
> app/build/intermediates/incremental/mergeDebugResources/merged.dir/values/values.xml:2222: warn: multiple substitutions specified in non-positional format; did you mean to add the formatted="false" attribute?.


tiembo:
Thanks!

PR #156, opened 2018-08-20
Update ktlint 0.27.0 description in README

XinyueZ:


tiembo:
Thank you for staying on top of version updates @XinyueZ !

PR #159, opened 2018-08-24
Update for gradle 3.2.0-rc02

XinyueZ:


XinyueZ:
Update gradle-3.2.0-rc01 -> rc02

PR #160, opened 2018-08-25
Move logical of "share detail" to PlantDetailViewModel

XinyueZ:
Don't expose biz-logical and data to Android's component.

- Including update of tests on PlantDetailViewModel.
- Check: https://developer.android.google.cn/topic/libraries/architecture/lifecycle#lc-bp
- Check: https://en.wikipedia.org/wiki/Single_responsibility_principle

@tiembo @andhie I have only moved operation of share to the view-model in order to be expected by practicing Jetpack documentation.


app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -82,28 +76,33 @@ class PlantDetailFragment : Fragment() {
...
                 return true
             }
             else -> super.onOptionsItemSelected(item)
         }
     }
+
+    @Suppress("DEPRECATION")
+    private fun onShare(shareText: String?) {
+        val text = shareText ?: ""
+        val shareIntent = ShareCompat.IntentBuilder.from(activity)
```
nic0lette: I'm curious why this code couldn't mostly be in the ViewModel as well? That is, something like: `private fun onShare(shareText: String?) { startActivity(viewModel.buildShareIntent(shareText)) }`

XinyueZ: The `view-model` should from the aspect of architecture know nothing about the `activity`.

hider-vass (approved):

hider-vass (approved):

PR #165, opened 2018-08-29
Update WorkManager to 1.0.0-alpha08

XinyueZ:


XinyueZ:
@tiembo I have tried the upgrade locally, it works fine, whatz wrong on CI?

tiembo:
@XinyueZ I think it was a temporary CI issue

PR #171, opened 2018-08-31
Fix lint warning in PlantListViewModel

XinyueZ:
Use `companion object` and `const` on `NO_GROW_ZONE` .


XinyueZ:
The error is strange, it works fine locally. 😭

tiembo:
Yeah, I'm not sure why this PR causes an unrelated test to fail.  Merging.

PR #179, opened 2018-09-11
feat: use showSnackbarMessage in viewmodel

williamwue:
Instead of show snackbar in fragment directly. Prepare for do remove garden planting action later (control delete action and show message action both in viewmodel).

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
@@ -22,6 +22,7 @@ import android.content.Intent
 import android.databinding.DataBindingUtil
 import android.os.Build
 import android.os.Bundle
+import android.support.annotation.StringRes
 import android.support.design.widget.Snackbar
 import android.support.v4.app.Fragment
 import android.support.v4.app.ShareCompat
@@ -34,6 +35,7 @@ import android.view.ViewGroup
 
 import com.google.samples.apps.sunflower.databinding.FragmentPlantDetailBinding
 import com.google.samples.apps.sunflower.utilities.InjectorUtils
+import com.google.samples.apps.sunflower.utilities.SnackbarMessage
 import com.google.samples.apps.sunflower.viewmodels.PlantDetailViewModel
 
 /**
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/google/samples/apps/sunflower/utilities/SingleLiveEvent.kt
```diff
@@ -0,0 +1,76 @@
+/*
```
XinyueZ: Use `EventObserver` pattern instead `SingleLiveEvent`
Check: http://tinyurl.com/y7e8hq4v
There's a blog link which describes how different between `EventObserver` & `SingleLiveEvent` which is important.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -60,10 +62,16 @@ class PlantDetailFragment : Fragment() {
             setLifecycleOwner(this@PlantDetailFragment)
             fab.setOnClickListener { view ->
```
XinyueZ: `view` which is local will never be used, use  `_` instead.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -60,10 +62,16 @@ class PlantDetailFragment : Fragment() {
...
                 plantDetailViewModel.addPlantToGarden()
-                Snackbar.make(view, R.string.added_plant_to_garden, Snackbar.LENGTH_LONG).show()
+                plantDetailViewModel.showSnackbarMessage(R.string.added_plant_to_garden)
```
XinyueZ: The Line 64-65 should happen asynchronously, that means the `message` would be shown just after `adding` behavior being caused which is incorrect information while the `inserting` is still running underway possibly.

**My opinion**  here is to use the **return** value(`LiveData<Long>`) of insert DAO which can be used by the view-model and show  `message` asynchronously.

app/src/main/java/com/google/samples/apps/sunflower/utilities/SingleLiveEvent.kt
williamwue: @XinyueZ OK, I'll learn it.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
williamwue: Got it.

mlykotom:
Thank you folks for the contribution, I'm going to close this as obsolete given the screen has been rewritten.

PR #183, opened 2018-09-13
Update ktlint to 0.28.0, it works now

XinyueZ:


tiembo:
Thanks for keeping the libs up to date! :+1:

XinyueZ:
🤝

PR #184, opened 2018-09-14
Bump versions of Kotlin and Support lib

tiembo:


tiembo:
Thank you @thagikura !

PR #190, opened 2018-09-20
Upgrade WorkManager to 1.0.0-alpha09 and update API calls

tiembo:


tiembo:
Offline LGTM from @pfmaggi

PR #194, opened 2018-09-20
For WorkManager, prefer using getApplicationContext() in doWork()

tiembo:


app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
```diff
@@ -27,18 +27,18 @@ import com.google.samples.apps.sunflower.data.AppDatabase
 import com.google.samples.apps.sunflower.data.Plant
 import com.google.samples.apps.sunflower.utilities.PLANT_DATA_FILENAME
 
-class SeedDatabaseWorker(val context: Context, workerParams: WorkerParameters) : Worker(context, workerParams) {
+class SeedDatabaseWorker(context: Context, workerParams: WorkerParameters) : Worker(context, workerParams) {
```
XinyueZ: `...:Worker(context.applicationContext, ...`, because,  ```SeedDatabaseWorker(context: Context,...``` might be called by an  `Activity` .

app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
```diff
@@ -27,18 +27,18 @@ import com.google.samples.apps.sunflower.data.AppDatabase
...
     private val TAG = SeedDatabaseWorker::class.java.simpleName
 
     override fun doWork(): Worker.Result {
         val plantType = object : TypeToken<List<Plant>>() {}.type
         var jsonReader: JsonReader? = null
 
         return try {
-            val inputStream = context.assets.open(PLANT_DATA_FILENAME)
+            val inputStream = applicationContext.assets.open(PLANT_DATA_FILENAME)
```
XinyueZ: The whole `return` should be oriented to `Kotlin` style:

```
  applicationContext.assets.open("").use {
            JsonReader(it.reader()).use {
                
            }
        }
```

app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
XinyueZ: @tiembo Please consider the PR #97 which has solved the similar issue there.

app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
tiembo: That's outside the scope of this PR

app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
tiembo: This is the preferred approach for Workers

XinyueZ: @tiembo 
Take a look at here
```
public abstract class NonBlockingWorker {

    @SuppressWarnings("NullableProblems")   // Set by internalInit
    private @NonNull Context mAppContext;
```
The grandfather of `Worker` has a member of  `Context`.  If `SeedDatabaseWorker` takes an `Activity` as `context` and passes to `Worker`, there could be a  `leak`.

app/src/main/java/com/google/samples/apps/sunflower/workers/SeedDatabaseWorker.kt
XinyueZ: ok, however, this approach can be a little more affluent to read than Java style.

pfmaggi: In some cases could be useful to have a local variable to avoid the call to getApplicationContext() (exposed by Kotlin as property: applicationContext). But in this case seems fair to just use the property syntax.

LGTM

XinyueZ: @pfmaggi My *changes requested*  here is about using `use` which can safe `close` the `closeable` interfaces like `stream` in order to reduce line like `finally{close()}`.

XinyueZ: The optimise #201 for this.

PR #196, opened 2018-09-22
Hero image transition

### LINKED ISSUE #66: Hero image transition animation
Doing transition animation to detail activity via the hero image

Work to do:
Use `ActivityOptionsCompat#makeSceneTransitionAnimation()`. for the animation by passing in activity and the imageview.
Each list item in Plant adapter should use `ViewCompat#setTransitionName()` and prefix/suffixed with item position. Both list and detail imageview need to have same transition name. this allows anim to transition backwards when user press back on Detail activity
We need to delay anim in detail via `supportPostponeEnterTransition()` until Glide loads the image and then use `supportStartPostponedEnterTransition()`

https://developer.android.com/reference/android/support/v4/app/ActivityOptionsCompat.html#makescenetransitionanimation_4
https://developer.android.com/reference/android/support/v4/view/ViewCompat.html#setTransitionName(android.view.View,%20java.lang.String)

andhie:
Fixes #66

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
@@ -19,7 +19,9 @@ package com.google.samples.apps.sunflower.adapters
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.databinding.DataBindingUtil
 import androidx.navigation.findNavController
+import androidx.navigation.fragment.FragmentNavigatorExtras
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import com.google.samples.apps.sunflower.PlantListFragment
@@ -35,8 +37,7 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
         val plant = getItem(position)
         holder.apply {
-            bind(createOnClickListener(plant.plantId), plant)
-            itemView.tag = plant
+            bind(createOnClickListener(), plant)
         }
     }
 
@@ -45,10 +46,15 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
                 LayoutInflater.from(parent.context), parent, false))
     }
 
-    private fun createOnClickListener(plantId: String): View.OnClickListener {
-        return View.OnClickListener {
-            val direction = PlantListFragmentDirections.ActionPlantListFragmentToPlantDetailFragment(plantId)
-            it.findNavController().navigate(direction)
+    private fun createOnClickListener(): OnPlantItemClickListener {
+        return object : OnPlantItemClickListener {
+            override fun onPlantItemClick(rootView: View, plant: Plant) {
+                val binding = DataBindingUtil.getBinding<ListItemPlantBinding>(rootView)
+                val navigatorExtras = FragmentNavigatorExtras(binding!!.plantItemImage to plant.plantId)
+
+                val direction = PlantListFragmentDirections.ActionPlantListFragmentToPlantDetailFragment(plant.plantId)
+                rootView.findNavController().navigate(direction, navigatorExtras)
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/google/samples/apps/sunflower/utilities/MoveViews.kt
```diff
@@ -0,0 +1,57 @@
+/*
```
XinyueZ: Duplicated copyright comments 😢

XinyueZ: Should remove one of them.

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
```diff
@@ -45,23 +46,32 @@ class PlantAdapter : ListAdapter<Plant, PlantAdapter.ViewHolder>(PlantDiffCallba
...
 
-    private fun createOnClickListener(plantId: String): View.OnClickListener {
-        return View.OnClickListener {
-            val direction = PlantListFragmentDirections.ActionPlantListFragmentToPlantDetailFragment(plantId)
-            it.findNavController().navigate(direction)
+    private fun createOnClickListener(): OnPlantItemClickListener {
+        return object : OnPlantItemClickListener {
+            override fun onPlantItemClick(rootView: View, plant: Plant) {
+                val binding = DataBindingUtil.getBinding<ListItemPlantBinding>(rootView)
+                val navigatorExtras = FragmentNavigatorExtras(binding!!.plantItemImage to plant.plantId)
```
XinyueZ: My view here that using `!!` here is really unsafe idea.

I'd like to do:

```
DataBindingUtil.getBinding<ListItemPlantBinding>(rootView)?.let{
    // blablaba
} ?: run {
   // Error-handling, like post a toast or snackbar events through LiveData.
}
```
The reason is that the click event is meanless when binding is really for some reasons `null` and error-handlings give fallback.

andhie: i did wrote the .let{} but it becomes too many lines just to generate a navigatorExtras. so i decide to let it be single line to be more concise and no reason it will be null.

do you have any suggestion how to use `ViewCompat.getTransitionName(view)`? the kotlin Pair expects a non-null String, but ViewCompats returns a nullable causing an error.

nic0lette:
Hi @andhie ! This is really cool! Thanks for working on it. I do have a few comments:

First, it looks like you've changed (slightly) click handling when adding the shared element transition. I wonder if you could revert that, so this PR is exclusively about the shared element transition.

If you slow the animation down (Developer options -> Transition animation scale + Animator duration scale, set both to 5x) you can see the transition feels a bit off. Specifically, while the image from the plant list moves to the header position smoothly, the fragment pops in with all the details first, and so the image looks like it slides over the fragment.

This is in contrast with the example animations from the Material Design website: https://material.io/design/components/lists.html#behavior -- The hero image should move and expand, either revealing the text as it expands or having the text fade in afterwards, so it looks as though the plant details fragment is unfolding from the plant item.

Does that make sense?

andhie:
i have revert the click listener code.
for the MD anim you mentioned, its the Condensed Transition type. it doesnt animate the hero image. which approach would be preferable? the one moving the hero image or the condensed transition
https://material.io/design/motion/customization.html#condensing-transitions

nic0lette:
Hi @andhie :) That makes sense. I think the animation that's on the right side (captioned with "The list item and focal element transition...") is close to what I was thinking about, and feels closer to what you had, since the hero image was animating up.

Does that help?

andhie:
@nic0lette i've tweak a little transition animation. 
there are areas i'm not satisfied with.

Enter Transition: the collapsingtoolbar title appears after everything completed
Return Transition: the text can still be seen for a brief moment during Fade

do you have any suggestions to improving this?

nic0lette:
Sorry for the slow reply :)

I took a look at this and I think the enter transition needs to fade the moving header image as it settles into place so that the collapsing toolbar (with the text) can be seen. (At least, what I *think* is happening is that the moving image is covering the toolbar and so, until it's removed at the end, the title isn't visible).

Unfortunately I'm still learning these things and so I wasn't able to quickly verify this, but you may be able to :)

andhie:
Changes in #315 makes the title appearing out of order is now invalid.

PR #203, opened 2018-09-25
Update Gradle wrapper to 4.10.2

friederbluemle:
This updates Gradle wrapper to the latest stable version `4.10.2`.
All files were updated automatically using the Gradle integrated `wrapper` task.

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the branch of this pull request. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest target branch. If it isn't, or if you'd like me to combine something, please let me know, and I will make the necessary updates as soon as possible. Thank you!</sup></sub>


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,22 +1,5 @@
-#
```
tiembo: Please leave this copyright header here.

friederbluemle:
@XinyueZ @tiembo - Thanks for your comments. As I mentioned in the PR description, the files were updated _automatically_ by the Gradle integrated wrapper script. I did not manually edit them.
If a _new_ Gradle wrapper is generated with a recent version of Gradle, you'll get the same updated shell scripts. The ones contained in this repo are just a leftover from an _old_ version of Gradle, and updating them somehow was missed when Gradle was updated at some point in the past (most likely by manually updating just the properties, instead of using the automated update task).
@tiembo - I readded the copyright comment by partially checking out `HEAD^` of my commit, and manually removing the deprecated timestamp.
Are there any tools available that add these copyright headers for you? It feels slightly inconvenient having to do this for a Gradle owned and generated file - and also: Is there anything in this autogenerated file that is "copyright-able"? ;) It is the same in all Gradle based repos.

friederbluemle:
Also rebased my commit.

XinyueZ:
@tiembo update of .gitignore is needed.😉
Expect wrapper, the other files should always being updated automatically in local environment.😋

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -14,9 +14,8 @@
 # limitations under the License.
 #
 
-#Mon Sep 11 11:17:48 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
XinyueZ: I prefer [gradle-4.10.2-bin.zip](https://stackoverflow.com/questions/25451308/android-studio-gradle-bin-zip-vs-gradle-all-zip)

friederbluemle: Why do you prefer gradle-4.10.2-bin.zip? The only thing it does it will save a few MB of download a _single time, machine wide_. On the other hand, you won't get the source code and documentation which can be useful when debugging Gradle scripts. Plus, the default setup for new projects generated with Android Studio is the `-all` variant, making it much more likely to be present on your machine already anyway, since most other projects will be using it.
Since this project is already using `-all` as well, I think it makes sense to just stick to it.

friederbluemle:
> I have checked on my local machine, there isn't any change on files of gradlew and gradle.bat.
Maybe you should avoid push these two file changes and push only the wrapper file. 😃

What's the reason you suggest to continue to use the outdated wrapper scripts?

Did you use the Gradle wrapper task to perform the update? You will see that it will update the scripts to the latest version:

    ./gradlew wrapper --gradle-version 4.10.2 --distribution-type all

XinyueZ:
@tiembo What do you think?

friederbluemle:
@tiembo I re-added the copyright header, could you check please?
Any other comments @tiembo @XinyueZ Thanks!

friederbluemle:
@tiembo @XinyueZ - I just rebased this PR onto the latest `master` branch. Any other comments? Thanks.

tiembo:
Hi @friederbluemle - thank you for your patience and the explanation about the updated wrapper - especially on how to update it for future versions.

friederbluemle:
Thanks @tiembo :bow:

PR #209, opened 2018-09-27
Update kotlin 1.3.0-rc-116

XinyueZ:
Check https://plugins.jetbrains.com/plugin/6954-kotlin


XinyueZ:
Again: The `test_shouldBeWatered` is always a problem in sometimes.  Need a ticket to fix this *test*.
I have just updated kotlin version, however, this test breaks down whole.

XinyueZ:
Yes, I'd like to fix it, please.

PR #219, opened 2018-10-08
Simplify checking for empty lists.

johnjohndoe:


app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
```diff
@@ -52,8 +52,7 @@ class GardenFragment : Fragment() {
         })
 
         viewModel.plantAndGardenPlantings.observe(viewLifecycleOwner, Observer { result ->
-            if (result != null && result.isNotEmpty())
-                adapter.submitList(result)
+            adapter.submitList(result.orEmpty())
```
XinyueZ: origin consumption: The logical operation 
after-change consumption:  Null or an empty list will continue to be called deep, consuming the runtime.

Take a look the internal what will happen:

```java
    public void submitList(@Nullable final List<T> newList) {
                     .......
              if (newList == null) {
                      mReadOnlyList = Collections.unmodifiableList(newList);
                      // notify last, after list is updated
                      mUpdateCallback.onInserted(0, newList.size());
                }
            .......
     }
```
I still prefer the origin which demonstrates to avoid submiting list clearly under conditations, although the `result.orEmpty()` is  a really comfortable method. 👍

tiembo:
Hi @johnjohndoe , thanks for the PR.  While `orEmpty()` is a useful method its use here would result in submitting an empty list to the adapter, as @XinyueZ mentioned.  In this case we only want to submit a list to the adapter then result is not null.

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
johnjohndoe: I expected the empty list to be an edge case which doesn't happen too often. I most of the cases a list of items would be passed to adapter. Further action can be limited by using DiffUtil.

PR #223, opened 2018-10-12
issue#111 Extract plant click listener logic from Adapter to Fragment

vjgarciag96:
As explained on issue 111, the logic of the plant item click listener was implemented on *PlantAdapter*. The new changes are about extracting this logic from the *PlantAdapter* to *PlantListFragment*. Instead of creating an interface and passing it to adapter for using it as callback, the approach implemented make use of a lambda function.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
@@ -26,6 +26,7 @@ import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProviders
+import androidx.navigation.findNavController
 import com.google.samples.apps.sunflower.adapters.PlantAdapter
 import com.google.samples.apps.sunflower.databinding.FragmentPlantListBinding
 import com.google.samples.apps.sunflower.utilities.InjectorUtils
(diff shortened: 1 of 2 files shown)


vjgarciag96:
I signed it!! 👍

app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
```diff
@@ -46,7 +47,11 @@ class PlantListFragment : Fragment() {
         val factory = InjectorUtils.providePlantListViewModelFactory(context)
         viewModel = ViewModelProviders.of(this, factory).get(PlantListViewModel::class.java)
 
-        val adapter = PlantAdapter()
+        val plantClickListener: (String) -> Unit = {
```
XinyueZ: nice 2 see, however, it doesn't change anything and bring benefits to the app, would you mind give some reasons why it should be extracted?

If I do, I'll create a `PlantItemViewModel` which impl. `OnClickListener` and holds `Plant`, meanwhile let each item of `RecyclerView` bind with the view-model, of cos, listener to `onClick` as well.

vjgarciag96: You are right, the changes of this branch only move the click listener logic from adapter to fragment and don't bring benefits to the app. I have the changes you are suggesting in another branch. I will open another Pull Request and we could discuss there about the new approach.

PR #226, opened 2018-10-12
issue#111 Create PlantItemViewModel for using it as binding on every plant list item instead of plant and its click listener

vjgarciag96:
The changes of this branch are about creating a view model for wrapping a plant and its click listener and use it as binding instead of having both the plant item and the click listener as bindings. Moreover, this changes let us to take out the logic of click action from adapter and use it only for delegating the user actions to collaborators (in this case the view model)


app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
```diff
@@ -70,7 +71,11 @@ class PlantListFragment : Fragment() {
 
     private fun subscribeUi(adapter: PlantAdapter) {
         viewModel.getPlants().observe(viewLifecycleOwner, Observer { plants ->
-            if (plants != null) adapter.submitList(plants)
+            if (plants != null) {
+                adapter.submitList(plants.map {
```
XinyueZ: The item view-model isnot created by `ViewModelProviders` , I do 
 prefe to `new` directly. 

Almost same case `PlantAndGardenPlantingsViewModel`. If you really want to let each item lifecycle sensitively, you can let all items `LifecycleObserver`. I use those pattern in MVVM-driven project.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantItemViewModel.kt
```diff
@@ -0,0 +1,19 @@
+package com.google.samples.apps.sunflower.viewmodels
+
+import android.view.View
+import androidx.lifecycle.ViewModel
+import androidx.navigation.findNavController
+import com.google.samples.apps.sunflower.PlantListFragmentDirections
+import com.google.samples.apps.sunflower.data.Plant
+
+class PlantItemViewModel(
+    private val plant: Plant
```
XinyueZ: Let public val is ok, no need to import a extra function to do same thing.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantItemViewModelFactory.kt
```diff
@@ -0,0 +1,13 @@
+package com.google.samples.apps.sunflower.viewmodels
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import com.google.samples.apps.sunflower.data.Plant
+
+class PlantItemViewModelFactory(private val plant: Plant) : ViewModelProvider.NewInstanceFactory() {
```
XinyueZ: For item of list, really find unnecessary.
Up 2 U. :smile_cat:

app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
vjgarciag96: Yes, it seems that having a factory for the view model of the item of a list is a bit useless. I have updated the code for removing the factoring and instantiating the view model directly.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantItemViewModelFactory.kt
vjgarciag96: Same as commented above:
> Yes, it seems that having a factory for the view model of the item of a list is a bit useless. I have updated the code for removing the factoring and instantiating the view model directly

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantItemViewModel.kt
vjgarciag96: ok, changed

app/src/main/java/com/google/samples/apps/sunflower/adapters/PlantAdapter.kt
```diff
@@ -17,26 +17,23 @@
...
+class PlantAdapter : ListAdapter<PlantItemViewModel, PlantAdapter.ViewHolder>(PlantDiffCallback()) {
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
-        val plant = getItem(position)
+        val plantViewModel = getItem(position)
         holder.apply {
-            bind(createOnClickListener(plant.plantId), plant)
-            itemView.tag = plant
+            bind(plantViewModel)
+            itemView.tag = plantViewModel.plant
```
XinyueZ: I think, now it is no more needed to use `tag` to point to `Plant` when we use view-model to encapsulate domain.

XinyueZ:
Would you mind to squash all commits into 99a2c439194bc9783f24919daf4a788b3a51880b which is the first commit of the whole story or second one which is the refinement of the first according to PR review, keep commits clean?

vjgarciag96:
> Would you mind to squash all commits into [99a2c43](https://github.com/googlesamples/android-sunflower/commit/99a2c439194bc9783f24919daf4a788b3a51880b) which is the first commit of the whole story or second one which is the refinement of the first according to PR review, keep commits clean?

They are all squashed into [e29828e](https://github.com/googlesamples/android-sunflower/pull/226/commits/e29828e20863bd5fe2e40400568105d3b9a6bd69). I don't know if it's valid to you.

XinyueZ:
@vjgarciag96 Actually and frankly I want to only keep commits clean and know how many main works are done, for example: feature done, and then refeinment .
Anyway, now the PR looks good and MVVM like. 👍

vjgarciag96:
@XinyueZ ok, if you want I can cherrypick the commit [e29828e](https://github.com/googlesamples/android-sunflower/pull/226/commits/e29828e20863bd5fe2e40400568105d3b9a6bd69) to another branch to make it cleaner. However, if you want to merge the current branch is also good for me.

XinyueZ:
@vjgarciag96 I think everything is fine. Because I always `rebase` after works being down,  i.e run `git rebase -i HEAD~7` to keep my commits cleaned, for your case like here `7` to squash all to first commit. Thatz only nice2have.

vjgarciag96:
@XinyueZ ok, perfect then. I didn't understand you at first, but know I understand it and sounds good. Thank you!

PR #229, opened 2018-10-19
Update kotlin-1.3.0-rc-190

XinyueZ:


tiembo:
I think a newer version is out?

XinyueZ:
@tiembo I have used the office to fetch available versions
https://plugins.jetbrains.com/plugin/6954-kotlin

tiembo:
@XinyueZ You're correct, I was thinking of another library.  Thanks!

PR #230, opened 2018-10-23
Update ktlint Installation Instructions

mrk-han:
This pull request updates the ktlint installation instructions to match the [README](https://github.com/shyiko/ktlint/blob/master/README.md) of [ktlint](https://github.com/shyiko/ktlint). 

Specifically, we add `&& sudo mv ktlint /home/local/bin` to  `curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint && chmod a+x ktlint`

Since ktlint was not in my `/home/local/bin` directory, on my setup of macOS using ZSH, I was unable to run `ktlint --apply-to-idea-project --android`

This PR adds installation instructions to avoid: `zsh: command not found: ktlint`

It also adds a blurb about installing via homebrew (also found from ktlint readme) which will by default put ktlint into `/home/local/bin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,9 @@ from the ktlint [README](https://github.com/shyiko/ktlint/blob/master/README.md)
 
 - Download ktlint:
 
-  `curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint && chmod a+x ktlint`
+  `curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/local/bin/`
+
+  On macOS (or [Linux](http://linuxbrew.sh/)) you can also use [brew](https://brew.sh/) - `brew install shyiko/ktlint/ktlint`
 
 - Inside the project root directory run:
 
--- a/README.md
+++ b/README.md
@@ -100,19 +100,11 @@ from the ktlint [README](https://github.com/shyiko/ktlint/blob/master/README.md)
 
 - Close Android Studio if it's open
 
-- Download ktlint:
-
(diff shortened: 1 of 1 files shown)


mrk-han:
I can see a benefit of only downloading ktlint to the project root directory. Perhaps my problem was zsh specific. If so, maybe we can add a line such as 

> "If using ZSH, you will have to download ktlint to `/home/local/bin`"

README.md
```diff
@@ -102,7 +102,9 @@ from the ktlint [README](https://github.com/shyiko/ktlint/blob/master/README.md)
 
 - Download ktlint:
 
-  `curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint && chmod a+x ktlint`
+  `curl -sSLO https://github.com/shyiko/ktlint/releases/download/0.29.0/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/local/bin/`
+
+  On macOS (or [Linux](http://linuxbrew.sh/)) you can also use [brew](https://brew.sh/) - `brew install shyiko/ktlint/ktlint`
```
XinyueZ: Is this "must" ?

tiembo:
Hi @mrk-han , thanks for the heads up and the PR for the updated ktlint instructions.

Since the installation instructions have expanded I prefer to link to the ktlint installation instructions rather than duplicate that text here, especially since it can vary based on OS.  How about replacing the "Download ktlint" bullet point with something like:

* Follow the [installation instructions to download ktlint](https://github.com/shyiko/ktlint/blob/master/README.md#installation)

README.md
mrk-han: I am actually not sure if it is something specific to ZSH or not, so for now I can only say that it is a "must" with my setup of using ZSH, though I am not above the idea that it may be user error. 

This is installation instruction is directly from the ktlint readme.  I was not able to run `ktlint --apply-to-idea-project --android` out of the box with the current instructions from the Sunflower Readme.

tiembo:
Looks good, thanks for contributing! :+1:

PR #247, opened 2018-11-16
Add non-goals section to README with specific note on DI

tiembo:


XinyueZ:
Actually, I am not one of the DI fans :)

PR #256, opened 2018-11-30
Delete feature

oussemaAr:
Adding garden's delete feature on long click on the item with a confirmation alert dialog

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
@@ -18,10 +18,13 @@ package com.google.samples.apps.sunflower.adapters
 
 import android.content.Context
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import androidx.databinding.DataBindingUtil
+import androidx.navigation.findNavController
 import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
+import com.google.samples.apps.sunflower.GardenFragmentDirections
 import com.google.samples.apps.sunflower.R
 import com.google.samples.apps.sunflower.data.PlantAndGardenPlantings
 import com.google.samples.apps.sunflower.databinding.ListItemGardenPlantingBinding
@@ -44,8 +47,16 @@ class GardenPlantingAdapter(
         getItem(position).let { plantings ->
             with(holder) {
                 itemView.tag = plantings
-                bind(plantings)
+                bind(createOnClickListener(plantings.plant.plantId), plantings)
             }
+        }
+    }
+
+    private fun createOnClickListener(plantId: String): View.OnClickListener {
+        return View.OnClickListener {
+                val direction =
+                        GardenFragmentDirections.ActionGardenFragmentToPlantDetailFragment(plantId)
+                it.findNavController().navigate(direction)
         }
     }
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
```diff
@@ -16,48 +16,97 @@
...
+import com.google.samples.apps.sunflower.viewmodels.PlantDetailViewModel
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Dispatchers.Main
+import kotlinx.coroutines.MainCoroutineDispatcher
+import kotlinx.coroutines.launch
 
 class GardenPlantingAdapter(
-    val context: Context
+        val context: Context
```
XinyueZ: This member is `no` need.

app/src/main/java/com/google/samples/apps/sunflower/data/GardenPlantingDao.kt
```diff
@@ -48,6 +48,6 @@ interface GardenPlantingDao {
     @Insert
     fun insertGardenPlanting(gardenPlanting: GardenPlanting): Long
 
-    @Delete
-    fun deleteGardenPlanting(gardenPlanting: GardenPlanting)
+    @Query("DELETE FROM garden_plantings WHERE plant_id = :plantId")
```
XinyueZ: I prefer to use  the origin .

app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
```diff
@@ -16,48 +16,97 @@
...
+        return View.OnClickListener {
+            val direction =
+                    GardenFragmentDirections.ActionGardenFragmentToPlantDetailFragment(plantId)
+            it.findNavController().navigate(direction)
+        }
+    }
+
+    private fun createOnLongClickListener(plantId: String): View.OnLongClickListener {
+        return View.OnLongClickListener {
+            AlertDialog.Builder(context, R.style.Base_ThemeOverlay_MaterialComponents_Dialog_Alert)
```
XinyueZ: The whole content of the listener should be handled with `LiveData`, the total clicking will be done in the `fragment`  logical. 
Due to create `Dialog` here, the member `context` should be an `Activity` variant which leaks all, even if `it.context` can be used, the approach to handle click in the adapter is not cool.

nic0lette:
Hi oussemaAr! Thanks for your PR!

I have a few reservations about the UX of this feature that I'm hoping you could address (outside of any coding implementation details that @XinyueZ mentioned).

It feels to me as though long pressing and getting a dialog to remove the plant isn't very common any longer. In fact, I can imagine two types of UX that would be better:

1) Long press to select plants, and then add a menu item (with icon) to delete the plants. Clicking on the 'delete' icon could pop-up a dialog, or could simply remove them with a short lived snackbar having an option to 'undo'. (GMail does this when selecting multiple messages to archive, delete, etc...)
2) Swipe the plant away, with a colored background under the plant item and a delete icon that's exposed as the user swipes, also similar to GMail. After the plant is removed, a snackbar presents the option to undo the action for a short time.

I can see benefits to both of them. Swiping away is more convenient, but selecting plants opens the way to enable other actions on single and multiple plants.

Unfortunately I'm just not particularly fond of the current flow. =(

Issue #272, opened 2018-12-12
Move click handling from Adapters


tiembo:

XinyueZ:
https://github.com/googlesamples/android-sunflower/pull/290#issuecomment-448146461

JoseAlcerreca:
(Use Event, not SingleLiveEvent :) )

XinyueZ:
@JoseAlcerreca Would you mind take a quick look at #291 ?

PR #290, opened 2018-12-16
[OPTIMIZE] #272: Move click handling from Adapters

XinyueZ:
Issue #272


tiembo:
Hi @XinyueZ , thanks for the PR.  I think we're taking the approach of a SingleLiveEvent ([article](https://medium.com/androiddevelopers/livedata-with-snackbar-navigation-and-other-events-the-singleliveevent-case-ac2622673150)), so these files will undergo a larger change beyond moving the handling logic to the ViewHolder.

I've updated #272 to clarify.

XinyueZ:
@tiembo The SingleLiveEvent has limit. Think about https://github.com/nickbutcher/plaid/tree/master/core/src/main/java/io/plaidapp/core/util/event

PR #291, opened 2018-12-18
[OPTIMIZE] #272: Move click handling from Adapters

XinyueZ:


app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
```diff
@@ -55,5 +56,11 @@ class GardenFragment : Fragment() {
             if (result != null && result.isNotEmpty())
                 adapter.submitList(result)
         })
+
+        adapter.openPlant.observe(viewLifecycleOwner, Observer { plantId ->
+            val direction =
+                GardenFragmentDirections.ActionGardenFragmentToPlantDetailFragment(plantId.peek())
```
nic0lette: Can you change these to use `plantId.consume()?.let { ... }`? We want to make sure that even if there are (somehow) multiple observers, only one handles the event. [Recommended: Use an Event wrapper](https://medium.com/androiddevelopers/livedata-with-snackbar-navigation-and-other-events-the-singleliveevent-case-ac2622673150#0e87)

app/src/main/java/com/google/samples/apps/sunflower/event/EventObserver.kt
```diff
@@ -0,0 +1,31 @@
+/*
```
nic0lette: This file seems to be unused?

XinyueZ: @nic0lette my mis, please check again.

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
XinyueZ: @nic0lette my mis, please check again.

JoseAlcerreca:
Hey @XinyueZ, what's the benefit of exposing the event from the adapter instead of using the navigation controller directly? 

Please always provide a description of the change.

XinyueZ:
@JoseAlcerreca Actually I am doing this according to the demand of #272 to move click content from adapter.  
But I prefer for #290  variant which still encapsulates navigation logical as part of viewholder.  My opinion is to treat the app-navigation simply, not to be a part of UI.

app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
```diff
@@ -21,19 +21,23 @@ import android.view.LayoutInflater
...
 import com.google.samples.apps.sunflower.databinding.ListItemGardenPlantingBinding
+import com.google.samples.apps.sunflower.event.Event
 import com.google.samples.apps.sunflower.viewmodels.PlantAndGardenPlantingsViewModel
 
 class GardenPlantingAdapter(
     val context: Context
 ) : ListAdapter<PlantAndGardenPlantings, GardenPlantingAdapter.ViewHolder>(GardenPlantDiffCallback()) {
 
+    private val _openPlant = MutableLiveData<Event<String>>()
+    internal val openPlant: LiveData<Event<String>> = _openPlant
```
mcastagnolo: I don't see why this has to be a LiveData vs a lambda or a Listener. I get that the Event type allows for one-time handled events, so this would operate similar to a lambda callback or Listener, and that there's an added benefit that a fragment/activity would only execute the attached logic if it was in an appropriate lifecycle state. But, is a click happening during a bad lifecycle state a real possibility here, can it not be handled with a lifecycle check by the owner, and is it too much complexity vs just setting a listener / using a lambda in the constructor?

XinyueZ: @mcastagnolo how about #290 ?

XinyueZ:
I have solved conflicts.

XinyueZ:
@nic0lette @tiembo any interest to take a look at this change?

arriolac:
Closing this as this is no longer relevant to the app since it's been migrated to Compose

PR #304, opened 2019-01-18
Omit .applicationContext in SeedDatabaseWorker

tiembo:
Per https://issuetracker.google.com/issues/123014564, this isn't necessary.


tiembo:
cc @XinyueZ

PR #305, opened 2019-01-18
dagger2 integration

trevorwang:


tiembo:
Hi @trevorwang , thank you for the detailed PR!  At this time, there are [no plans] to incorporate dependency injection into Sunflower, as to help developers unfamiliar with DI better understand Sunflower's code.

XinyueZ:
I prefer koin instead dagger if DI would be used which is more simple. 😄

PR #306, opened 2019-01-20
[UPDATE] Gradle distribution 5.1.1

XinyueZ:
- It works in AS 3.3
- Update associated files as well


gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
nic0lette: super nit: It looks like this got double quote enclosed.... That is, I think you can drop the `'` marks and leave only the `"` marks.

XinyueZ: I did not do this by my hand.
Try this `./gradlew wrapper --gradle-version 5.1.1 --distribution-type all`

nic0lette: You're right. Oops! Thanks.

PR #311, opened 2019-01-24
Using kotlin method to make a null validate

MathBrandino:
don't use if validation cause like java and Kotlin is more clean

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
@@ -70,7 +70,7 @@ class PlantListFragment : Fragment() {
 
     private fun subscribeUi(adapter: PlantAdapter) {
         viewModel.getPlants().observe(viewLifecycleOwner, Observer { plants ->
-            if (plants != null) adapter.submitList(plants)
+            plants?.let{ adapter.submitList(it) }
         })
     }
 


app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt
```diff
@@ -70,7 +70,7 @@ class PlantListFragment : Fragment() {
 
     private fun subscribeUi(adapter: PlantAdapter) {
         viewModel.getPlants().observe(viewLifecycleOwner, Observer { plants ->
-            if (plants != null) adapter.submitList(plants)
+            plants?.let{ adapter.submitList(it) }
```
XinyueZ: I prefer `      if (plants != null && plants.isNotEmpty()) adapter.submitList(plants)`

andhie: even if the list is empty, i think it should be submitted to the adapter.

XinyueZ: @andhie not so perfect. take a look a source code, you'll see that there is some complex logic which will be processed.

nic0lette: Really? I just took a look, and now I wonder what I'm missing:

With `if`:
```
         public final void onChanged(List plants) {
            if (plants != null) {
               adapter.submitList(plants);
            }
         }
```

With `let`:
```
         public final void onChanged(List plants) {
            if (plants != null) {
               int var4 = false; /* ??? */
               adapter.submitList(plants);
            }
         }
```

The only other difference I see is the 'DUP' to create `it` which we don't use and I *hope* gets optimized farther down the toolchain.

XinyueZ: @nic0lette 😄 I mean here:
The source of  `submitList` 
```
public void submitList(@Nullable final List<T> newList) {
        final int runGeneration = ++this.mMaxScheduledGeneration;
        if (newList != this.mList) {
            if (newList == null) {
                int countRemoved = this.mList.size();
                this.mList = null;
                this.mReadOnlyList = Collections.emptyList();
                this.mUpdateCallback.onRemoved(0, countRemoved);
            } else if (this.mList == null) {
                this.mList = newList;
                this.mReadOnlyList = Collections.unmodifiableList(newList);
                this.mUpdateCallback.onInserted(0, newList.size());
            } else {
                final List<T> oldList = this.mList;
                this.mConfig.getBackgroundThreadExecutor().execute(
```

andhie: i will still prefer to cater a case where have list -> no list to be diff/animated.

XinyueZ: People can write an extension to wrapper the logic and the codes will be looked gracefully as well

PR #315, opened 2019-01-31
Fix double toolbar issue in Plant Details page

tiembo:
Changes include:
- Remove the duplicate toolbar in the collapsing layout
- Show plant name above watering details
- Change text size of plant description

We will look into a different layout when we start implementing Material design.

### DIFF
--- a/app/src/main/res/layout/fragment_plant_detail.xml
+++ b/app/src/main/res/layout/fragment_plant_detail.xml
@@ -59,13 +59,6 @@
                     app:imageFromUrl="@{viewModel.plant.imageUrl}"
                     app:layout_collapseMode="parallax" />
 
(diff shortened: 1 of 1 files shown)


tiembo:
Fixes #75

app/src/main/res/layout/fragment_plant_detail.xml
```diff
@@ -83,21 +76,35 @@
                 android:layout_height="match_parent">
 
                 <TextView
-                    android:id="@+id/plant_watering"
+                    android:id="@+id/plant_name"
+                    style="?android:attr/textAppearanceLarge"
```
andhie: Using android:textAppearance would be more suitable than using style as this attr only affects text display

Issue #278, opened 2018-12-12
Dark mode


tiembo:

nic0lette:
My initial thought is that it would be a setting, rather than being dependent upon time of day, since I know some would prefer to use a dark theme all of the time, and others would rather stick with a light theme at all times.

XinyueZ:
@tiembo @nic0lette If you dont mind I want to know the color of following ones
```xml
   <color name="colorPrimary"></color>
    <color name="colorPrimaryDark"></color>
    <color name="colorAccent"></color>
```

tiembo:
This is partially implemented in the Material Spring changes.  Next is adding a setting in the app to choose to use light or dark theme in the app, overriding the system setting.

PR #320, opened 2019-02-07
[UPDATE] Material Design Lib 1.1.0-alpha02

XinyueZ:
Update the material design library to `1.1.0-alpha02` for the reasons:

`MD-Lib + Theme.MaterialComponents.DayNight.NoActionBar` .
One of preparations for #278


tiembo:
Hi @XinyueZ , thanks for pointing this out.  I think we should include the dependency upgrade the main PR for #278.  Otherwise we'll need to continually update this dependency even though the new alpha features aren't yet being used in the code.  Does that sound OK with you?

XinyueZ:
OK, @tiembo it looks not too bad, only here to give image that if people want to do night-light with MD(not support lib), they should use the MD alpha.

Issue #322, opened 2019-02-08
Add Japanese strings translation


tiembo:

XinyueZ:
I can, if you want to assign this ticket to me :)

tiembo:
Thanks for the PR @XinyueZ !

PR #328, opened 2019-02-10
[FEATURE] Japanese support for issue #322

XinyueZ:


app/src/main/res/values-jpn/strings.xml
```diff
@@ -0,0 +1,47 @@
...
+    <string name="garden_empty">あなたの庭に植物はありません</string>
+    <string name="planted_date">%1$s は %2$s に植わっています</string>
+    <string name="menu_item_share_plant">共有</string>
+    <string name="share_text_plant">%s という植物を「Sunflower」アプリで見てみてください</string>
+
+    <!-- String plurals and their related prefix / suffix strings -->
+    <!-- Item quantities for zero and one are not used in Chinese -->
+    <string name="watering_needs_prefix">水の必要量:</string>
+    <plurals name="watering_needs_suffix">
+        <item quantity="one">毎日</item>
```
nic0lette: I'm not sure Japanese will use "one" strings.

app/src/main/res/values-jpn/strings.xml
```diff
@@ -0,0 +1,47 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
+    <string name="menu_filter_by_grow_zone">成長度合いによって分けられる</string>
```
thagikura: If it's a label for a menu I think "成長度合いによって分ける" makes more sense

nic0lette: My Japanese is really _really_ bad, but I wonder about this one? That is, I wonder if something like 気候区分 might be better than 成長度合い because it's more about what climate the plant survives/thrives?

app/src/main/res/values-jpn/strings.xml
```diff
@@ -0,0 +1,47 @@
...
+    <!-- Item quantities for zero and one are not used in Chinese -->
+    <string name="watering_needs_prefix">水の必要量:</string>
+    <plurals name="watering_needs_suffix">
+        <item quantity="one">毎日</item>
+        <item quantity="other">毎 %d 日</item>
+    </plurals>
+
+    <string name="watering_next_prefix">最後に水を上げたのは: %s</string>
+    <plurals name="watering_next_suffix">
+        <item quantity="one">明日水をあげてください</item>
```
nic0lette: I don't think Japanese will use "one"

app/src/main/res/values-jpn/strings.xml
thagikura: Right. I think quantity="other" is always used for Japanese. Also "毎%d日" looks more natural (remove the spaces before and after %d)

tiembo:
Echoing @nic0lette - this should be in `values-ja` (similar to [iosched](https://github.com/google/iosched/blob/master/mobile/src/main/res/values-ja/strings.xml))

app/src/main/res/values-jpn/strings.xml
XinyueZ: done

app/src/main/res/values-jpn/strings.xml
XinyueZ: yes, looks better.

app/src/main/res/values-jpn/strings.xml
XinyueZ: fine, looks better.

tiembo:
Thank you @XinyueZ !

PR #329, opened 2019-02-11
Reduced animation time between fragment transitions

asadmansr:
Implemented a fix for issue #265. The reduced animation time provides a seamless transition between fragments.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ buildscript {
         // App dependencies
         constraintLayoutVersion = '2.0.0-alpha3'
         coreTestingVersion = '2.0.0'
-        coroutinesVersion = "1.0.1"
+        coroutinesVersion = "1.1.1"
         espressoVersion = '3.1.0-alpha4'
         glideVersion = '4.8.0'
         gradleVersion = '3.3.1'
--- a/app/src/main/res/anim/slide_in_right.xml
+++ b/app/src/main/res/anim/slide_in_right.xml
@@ -18,5 +18,5 @@
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <translate android:fromXDelta="100%" android:toXDelta="0%"
         android:fromYDelta="0%" android:toYDelta="0%"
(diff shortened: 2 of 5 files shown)


app/src/main/res/anim/slide_in_left.xml
```diff
@@ -18,5 +18,5 @@
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <translate android:fromXDelta="-100%" android:toXDelta="0%"
         android:fromYDelta="0%" android:toYDelta="0%"
-        android:duration="700"/>
+        android:duration="400"/>
```
XinyueZ: Would you mind define the `duration` in a resource file and people can change it just in one place later?

tiembo:
Thanks for the PR!  Could you rebase your PR?

I extracted out the values in #330 (thanks for the heads up @XinyueZ), so this should now be a one-line change.

asadmansr:
Thanks @XinyueZ and @tiembo for the review. I've updated the anim.xml file with the duration value.  So in the future, only one value will need to be changed if the duration of the animation changes.

I've also signed the CLA. Let me know if there's anything on my end to resolve the CLA check.

tiembo:
I think this PR might be in a weird state with the CLA.  Since it's just a one-line change, would you mind opening a new PR for this change?  It might be faster than trying to figure out what's going on with the CLA

build.gradle
```diff
@@ -25,7 +25,7 @@ buildscript {
         // App dependencies
         constraintLayoutVersion = '2.0.0-alpha3'
         coreTestingVersion = '2.0.0'
-        coroutinesVersion = "1.0.1"
+        coroutinesVersion = "1.1.1"
```
XinyueZ: 😃 An inconsistent change here, the change has been updated in upstream.
I suggest a new PR as well or rebase technical❓

asadmansr:
Sorry about that! I opened a new PR #333 with the one-line change. I'm closing this PR to avoid any confusions.

Issue #271, opened 2018-12-12
Add non-changing plant IDs


tiembo:
This lays the groundwork for updating the plant database without having to recreate the database or uninstalling and reinstalling the app.

(Update 2019-02-12) Possible candidate for uniquely classifying a plant is to use [Binomial nomenclature](https://en.wikipedia.org/wiki/Binomial_nomenclature) (such as ["helianthus annus"](https://en.wikipedia.org/wiki/Helianthus_annuus) for sunflower))

PR #332, opened 2019-02-12
[OPTIMIZE] Change plantId of class Plant to long

XinyueZ:
I think this was a bug that the `Plant`  has the `plantId: String`  but in the `GardenPlanting` which associates with `Plant` is using `foreignKeys` : `plantId: Long`

The commit has only changed `plantId: String` to `plantId: Long`, nothing else.

@tiembo @nic0lette

### DIFF
--- a/app/src/androidTest/java/com/google/samples/apps/sunflower/data/PlantDaoTest.kt
+++ b/app/src/androidTest/java/com/google/samples/apps/sunflower/data/PlantDaoTest.kt
@@ -33,9 +33,9 @@ import org.junit.runner.RunWith
 class PlantDaoTest {
     private lateinit var database: AppDatabase
     private lateinit var plantDao: PlantDao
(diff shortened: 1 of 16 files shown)


tiembo:
Hi @XinyueZ , we want to keep the plantId as a string, as we probably want to use binomial nomenclature for the unique plant ID (see updated comment on #271 )

XinyueZ:
@tiembo but why is it in `GardenPlanting` a `Long`, any reason?

tiembo:
@XinyueZ do you mean the primary key of `GardenPlanting`?  Since it is [autogenerated](https://developer.android.com/reference/androidx/room/PrimaryKey#autoGenerate()) it should be int or long.

XinyueZ:
ok @tiembo .when you sure you want to let them in differet types, I will close this ticket with no-change.

fine?

tiembo:
Yes, `Plant` and `GardenPlanting` have two different types for their primary keys.

PR #333, opened 2019-02-12
Reduced duration time of fragment transitions

asadmansr:
New PR for the implemented fix for issue #265. The reduced animation time provides a seamless transition between fragments. This also includes the prep work from #330

### DIFF
--- a/app/src/main/res/values/anim.xml
+++ b/app/src/main/res/values/anim.xml
@@ -19,5 +19,5 @@
 <resources>
 
     <!-- Duration (in milliseconds) for slide transition animation -->
-    <integer name="slide">700</integer>
+    <integer name="slide">400</integer>
 </resources>


app/src/main/res/values/anim.xml
```diff
@@ -19,5 +19,5 @@
 <resources>
 
     <!-- Duration (in milliseconds) for slide transition animation -->
-    <integer name="slide">700</integer>
+    <integer name="slide">400</integer>
```
XinyueZ: I don't want to hinder your thoughts, but I also recommend the following:
screen shot 2019-02-12 at 22 28 35

asadmansr: Oh, it seems like config_longAnimTime also equals to 400. In my opinion, having our own set value might enhance readability and flexibility if the duration requirement changes in the future. But I can update the PR, if this is required by all the reviewers.

nic0lette: Hm... I wonder if using `@android:integer/config_mediumAnimTime` might be a good choice then?

tiembo: Thanks for point that out, @XinyueZ.  Note that these config values may change with OS releases.  So depending on your needs it may make sense to match the framework's animation values.

For our purposes let's keep it at 400.

tiembo:
Thanks @asadmansr for creating a new PR for this change!

PR #348, opened 2019-02-21
[UPDATE/OPTIMISE] Introduce viewmodel-ktx and simplify PlantDetailViewMode

XinyueZ:
This PR based on #347, please check #347 firstly and then consinder this one.

@tiembo @nic0lette


nic0lette:
Awesome. Thank you!

PR #358, opened 2019-03-06
Update Navigation to 2.0.0-rc02

tiembo:


build.gradle
```diff
@@ -36,7 +36,8 @@ buildscript {
         ktxVersion = '1.0.1'
         lifecycleVersion = '2.1.0-alpha02'
         materialVersion = '1.0.0'
-        navigationVersion = '1.0.0-rc02'
+        navigationVersion = '2.0.0-rc02'
+        navigationSafeArgsVersion = '1.0.0-rc02'
```
ianhanniballake: Navigation and Safe Args should always align in version, so I think it makes sense to collapse these into just a single version number.

build.gradle
```diff
@@ -53,7 +54,7 @@ buildscript {
     dependencies {
         classpath "com.android.tools.build:gradle:$gradleVersion"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-        classpath "android.arch.navigation:navigation-safe-args-gradle-plugin:$navigationVersion"
+        classpath "android.arch.navigation:navigation-safe-args-gradle-plugin:$navigationSafeArgsVersion"
```
ianhanniballake: This should be androidx.navigation

build.gradle
tiembo: Done

build.gradle
tiembo: Done

Issue #362, opened 2019-03-09
[BUG] androidTest(connectedCheck) does not work more 


XinyueZ:
Cannot run .`/gradlew check connectedCheck ` successfully.

tiembo:
Fixed via #363

PR #363, opened 2019-03-09
[FIX] #362, connectedCheck (androidTest) broken

XinyueZ:


tiembo:
Thanks for fixing the test!  Opened #367 to have a more real-world navigation scenario for the test setup.

PR #365, opened 2019-03-09
[OPTIMIZE] Remove call on `tag` in adapters

XinyueZ:


nic0lette:
LGTM, but let me just ping @tiembo to see if there was any reason the tags were added.

XinyueZ:
@tiembo fine ?

PR #369, opened 2019-03-12
Nb/rv selection

nic0lette:


nic0lette:
This is actually just a merge from master -> nb/rv-selection to get that one back mostly in sync with other PRs before the next steps.

PR #371, opened 2019-03-12
Add ActionMode framework.

nic0lette:
Change-Id: I818485adb8783f6707f30b892375ba00410e578d

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
@@ -102,10 +102,7 @@ class GardenFragment : Fragment() {
             }
             actionMode?.title =
                 resources.getQuantityString(
-                    R.plurals.garden_action_mode_count,
-                    /* quantity= */ selectedCount,
-                    /* parameter value */ selectedCount
-                )
+                        R.plurals.garden_action_mode_count, selectedCount, selectedCount)
         }
     }
 
--- a/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
@@ -18,9 +18,12 @@ package com.google.samples.apps.sunflower
 
 import android.os.Bundle
 import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
-import androidx.activity.OnBackPressedCallback
+import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.view.ActionMode
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProviders
@@ -36,6 +39,8 @@ import com.google.samples.apps.sunflower.viewmodels.GardenPlantingListViewModel
 
 class GardenFragment : Fragment() {
 
+    private var actionMode: ActionMode? = null
+
     private lateinit var selectionTracker: SelectionTracker<Long>
     private lateinit var selectionObserver: GardenSelectionObserver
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
```diff
@@ -92,7 +89,24 @@ class GardenFragment : Fragment() {
...
 
     private fun onSelectionChanged(selectedCount: Int) {
-        // TODO: Enable/disable ActionMode when items are selected.
+        if (selectedCount == 0) {
+            actionMode?.let {
+                it.finish()
+                actionMode = null
+            }
+        } else {
+            (requireActivity() as? AppCompatActivity)?.let { activity ->
```
XinyueZ: I don't think that `requireActivity()` returns `nullable`.
(Because of `nullable` by `getActivity()`, Google provides `requireActivity()` which doesn't return `null`, as I know.)

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
```diff
@@ -101,23 +115,36 @@ class GardenFragment : Fragment() {
...
-            if (selectionTracker.hasSelection()) {
-                selectionTracker.clearSelection()
-                true
-            } else {
-                false
-            }
+    /**
+     * Inner class to encapsulate handling of the [ActionMode] in my garden.
+     */
+    private inner class GardenActionModeCallback : ActionMode.Callback {
```
XinyueZ: Take care, `inner class` has an implicit pointer to parent class(Fragment).

I do not know that Google uses for a standard solution to face the `inner class` (like java inner class), however, in my recent project, I did like this, which could be a bit of advice here:

```kotlin

inner class Foo(lifecycleOwner: LifecycleOwner) : LifecycleObserver {
        private var lifecycleOwner: LifecycleOwner? = lifecycleOwner
  
        @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
        fun onDestroy() {
            lifecycleOwner?.lifecycle?.removeObserver(this)
            lifecycleOwner = null
        }

        fun foo() = lifecycleOwner?.let {
                //TODO with foo task here.
         }
}

```

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
nic0lette: You're right, it's not `nullable`, but it returns a `FragmentActivity`, and we need an `AppCompatActivity`. The `as?` does the, try to cast this to an `AppCompatActivity`. If it *can't* (I don't know why it wouldn't, really), then, rather than throwing, make the value `null`. :)

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
nic0lette: Sure, but the `inner class` isn't passed anywhere, and so, while the fragment and inner class point at one another, nothing will prevent them both from being freed when the Fragment is released.

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
tiembo: :point_up: that's a great explanation.  Add it as a comment?

app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
```diff
@@ -92,7 +89,24 @@ class GardenFragment : Fragment() {
...
+        } else {
+            (requireActivity() as? AppCompatActivity)?.let { activity ->
+                if (actionMode == null) {
+                    actionMode = activity.startSupportActionMode(GardenActionModeCallback())
+                }
+            }
+            actionMode?.title =
+                resources.getQuantityString(
+                    R.plurals.garden_action_mode_count,
+                    /* quantity= */ selectedCount,
```
tiembo: I think it's ok to omit the param comments (they're not mentioned during other usages of `getQuantityString`)

nic0lette: Done.

PR #378, opened 2019-03-22
Use ArchTaskExecutor instead of InstantTaskExecutorRule

skydoves:
### Use ArchTaskExecutor instead of InstantTaskExecutorRule

```kotlin
@get:Rule
var instantTaskExecutorRule = InstantTaskExecutorRule()
```

The issue is that this one does not provide a real thread for main thread and accepts all threads as main; hence might hide some bugs by having a different behavior than the real implementation. (and it is likely to be deprecated - removed). by __yigit__.

#### Modified:
- removed InstantTaskExecutorRule
- added TaskExecutorUtils : registerTaskExecutor(), unRegisterTaskExecutor()

### DIFF
--- a/app/src/androidTest/java/com/google/samples/apps/sunflower/data/GardenPlantingDaoTest.kt
+++ b/app/src/androidTest/java/com/google/samples/apps/sunflower/data/GardenPlantingDaoTest.kt
@@ -16,29 +16,26 @@
 
 package com.google.samples.apps.sunflower.data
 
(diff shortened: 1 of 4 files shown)


XinyueZ:
@skydoves  I think Google has updated the test framework internal already, could you check?

XinyueZ:
@tiembo I think the newest google's repo has updated corresponded APIs.

PR #396, opened 2019-04-25
Nb/rv selection

nic0lette:


nic0lette:
Merged and pushed via merge on the CL.

PR #400, opened 2019-04-26
[FIX] Warning on call deprecated WorkManager.getInstance()

XinyueZ:
Call `WorkManager.getInstance(context)` instead.


tiembo:
Thanks!

PR #406, opened 2019-05-11
Initialize the ViewModel using the extension function

madroidmaq:
I replaced `ViewModelProviders.of(lifecycleOwner, factory).get(class)` with the `vimeModels` extension in `fragment-ktx`, which seems to be more in line with the `Kotlin` syntax, like this:

**From** 

```kotlin
val factory = InjectorUtils.provideGardenPlantingListViewModelFactory(requireContext())
val viewModel = ViewModelProviders.of(this, factory).get(GardenPlantingListViewModel::class.java)
```

**To**

```kotlin
private val viewModel: GardenPlantingListViewModel by viewModels {
        InjectorUtils.provideGardenPlantingListViewModelFactory(requireContext())
    }
```

Official documents can be viewed [here](https://developer.android.com/reference/kotlin/androidx/fragment/app/package-summary#viewmodels) .


nic0lette:
This is awesome! Thanks for the PR!

PR #407, opened 2019-05-14
Oberve LiveData with  Extension function

madroidmaq:
I used the extension function `observe` in `LiveData` to make adjustments to the code in the project, which seems to be more in line with the `Kotlin` syntax, like this:

**From** 

```kotlin
viewModel.gardenPlantings.observe(viewLifecycleOwner, Observer { plantings ->
    binding.hasPlantings = !plantings.isNullOrEmpty()
})
```

**To**

```kotlin
viewModel.gardenPlantings.observe(viewLifecycleOwner) { plantings ->
    binding.hasPlantings = !plantings.isNullOrEmpty()
}
```

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/GardenFragment.kt
@@ -22,7 +22,7 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.viewModels
-import androidx.lifecycle.Observer
+import androidx.lifecycle.observe
 import com.google.samples.apps.sunflower.adapters.GardenPlantingAdapter
 import com.google.samples.apps.sunflower.databinding.FragmentGardenBinding
 import com.google.samples.apps.sunflower.utilities.InjectorUtils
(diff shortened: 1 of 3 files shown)


madroidmaq:
What is the meaning of LGTM?

pfmaggi:
LGTM stands for Looks Good To Me

madroidmaq:
thanks

nic0lette:
Generally looks great! Can you clean up the minor things with spotless and then we can merge this?

madroidmaq:
@piotrek1543 You are right. 
`plants != null` means `plants` is nullable, there is no need to add a comment, but `Android Studio` will think `plants != null` is not necessary, like this:

image

This is because the Room database is declared to be non-nullable, like this:

```kotlin
    @Query("SELECT * FROM plants ORDER BY name")
    fun getPlants(): LiveData<List<Plant>>
```
You can find his definition [here](https://github.com/googlesamples/android-sunflower/blob/master/app/src/main/java/com/google/samples/apps/sunflower/data/PlantDao.kt). I think it should be defined as follows:

```kotlin
    @Query("SELECT * FROM plants ORDER BY name")
    fun getPlants(): LiveData<List<Plant>?>
```

Form `LiveData<List<Plant>>` To `LiveData<List<Plant>?>`, From notnull to nullable.

You can find the implementation class `PlantDao_Impl.java` generated by Room in the `app/build/generated/source/kapt/debug/[package_name]/data/PlantDao_Impl.java` directory.

tiembo:
Thank you @madroidmaq !

For the `.observe` extension function issue, can you file an issue at https://developer.android.com/jetpack/docs/feedback?

Issue #419, opened 2019-06-03
README Support section refers to a Google+ community


gryclmn:
The README Support section links to a Google+ Community. Since Google+ is shut down should this be removed?

tiembo:
Yep - thanks for submitting the PR!

PR #421, opened 2019-06-03
Removed Google+ community reference since Google+ is shutdown.

gryclmn:
This PR is regarding issue #419.

Removed reference and link to Google+ Community since Google+ is shutdown.


tiembo:
Thank you!

PR #424, opened 2019-06-04
Ref PlantDao to return LiveData type, change to nullable type

madroidmaq:
According to the SQL statement, the query from the database by id does not necessarily return a value. Therefore, the generics in LiveData should be nullable.

This is also confirmed in the code(`PlantDao_Impl`) generated by the Room, as follows:

image

In addition, in the `PlantDetailFragment` class will also determine whether the plant is null, as follows:

image


madroidmaq:
@XinyueZ  Sure, I like to do this. I think the commit should be atomized as much as possible, so I did this in another commit(#426).

XinyueZ:
> @XinyueZ Sure, I like to do this. I think the commit should be atomized as much as possible, so I did this in another commit(#426).

🤝

PR #426, opened 2019-06-05
Using let instead if condi check

madroidmaq:
using `let` instead `if condi` check, you  can see more details here(#424).


nic0lette:
This looks good, but I wonder, what if `shareText` was something more like:

```
private val shareText
    get() = plantDetailViewModel.plant.value?.let { plant ->
        getString(R.string.share_text_plant, plant.name)
    } ?: ""
```

This would remove the need to have the observer, and it seems like the string would be correct when it's needed.

What are your thoughts, @XinyueZ (Ping @tiembo :)

tiembo:
Spoke with @nic0lette offline - let's give the `get()` approach a try.

madroidmaq:
The way to use `get()` is really elegant, I have made some adjustments.

ImYoungUn:
@madroidmaq  @XinyueZ   I'm sorry to be a novice at the "pull request" action, but can you tell me why this pool request is not merged?

PR #433, opened 2019-06-17
Change Plant List Page to use GridLayout

BatmanInGotham2:
Changed XML file of Plant List fragment to use GridLayout Manager instead of LinearLayoutManager, and changed GridLayout to use two columns instead of one. This is a setup PR to update the Plant List page to use Material Design cards.

Screen Shot 2019-06-17 at 2 55 22 PM

Screen Shot 2019-06-17 at 2 55 36 PM

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -34,4 +34,7 @@
     <dimen name="navigation_drawer_header_height">192dp</dimen>
     <dimen name="navigation_drawer_header_padding">16dp</dimen>
     <dimen name="navigation_drawer_image_width">100dp</dimen>
+
+    <!-- Number of columns for garden and plant list grid -->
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/dimens.xml
```diff
@@ -34,4 +34,7 @@
     <dimen name="navigation_drawer_header_height">192dp</dimen>
     <dimen name="navigation_drawer_header_padding">16dp</dimen>
     <dimen name="navigation_drawer_image_width">100dp</dimen>
+
+    <!-- Number of columns for garden and plant list grid -->
+    <integer name="grid_columns">2</integer>
```
YassinAJDI: Why not move `<integer name="grid_columns">2</integer>` inside `res/integers.xml`

The `dimens.xml` file should only contain dimension values with a unit of measure.

PR #452, opened 2019-06-23
[UPDATE]  work -> 2.1.0beta02

XinyueZ:


nic0lette:
Sorry, I think this fell through the cracks during the redesign and wasn't ever merged. I *think* we could bump work manager to 2.2.0, but I haven't done any testing with that.

In either case, I'll just close this for now since the project is already using the stable version of 2.1.0.

PR #464, opened 2019-06-26
fixed padding in plant list page

BatmanInGotham2:
Fixed side paddings for recycler view and fixed header padding.

Screenshot_1561583689


app/src/main/res/layout/plant_list_header.xml
```diff
@@ -24,7 +24,9 @@
         android:id="@+id/header_plant_list"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:layout_margin="@dimen/header_margin"
+        android:layout_marginStart="@dimen/card_side_margin"
```
tiembo: Is `marginEnd` also needed?

PR #468, opened 2019-06-28
Add icons to the tab layout

tiembo:
The icons are programmatically set based on the selected tab.

Screenshot_1561759670


app/src/main/res/drawable/ic_my_garden_active.xml
```diff
@@ -0,0 +1,22 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<vector android:height="24dp" android:viewportHeight="21"
+    android:viewportWidth="18" android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="#FFFF63" android:fillType="nonZero"
```
BatmanInGotham2: might be better to use ?attr/ for colors in case color pallette changes

tiembo: Done

PR #482, opened 2019-07-10
Removed use of .tag and Renamed ViewPager

BatmanInGotham2:


app/src/main/java/com/google/samples/apps/sunflower/HomeViewPagerFragment.kt
```diff
@@ -61,4 +61,13 @@ class ViewPagerFragment : Fragment() {
 
         return binding.root
     }
+
+    /** Return the list of tab titles to pass to SunflowerPagerAdapter **/
+    private fun getTabTitles(): List<String> {
+        val tabTitles: MutableList<String> = arrayListOf()
```
tiembo: What do you think about using a HashMap?  That way the case statement in `getPageTitle` can be simplified to `tabTitles.get(position)`?

BatmanInGotham2: I agree, it would make for faster indexing too

BatmanInGotham2: updated

app/src/main/java/com/google/samples/apps/sunflower/adapters/SunflowerPagerAdapter.kt
```diff
@@ -38,18 +38,10 @@ class SunflowerPagerAdapter(
...
+        return tabFragments[position] ?: throw IndexOutOfBoundsException()
     }
 
     override fun getPageTitle(position: Int): CharSequence? {
-        return when (position) {
-            MY_GARDEN_PAGE_INDEX -> context.getString(R.string.my_garden_title)
-            PLANT_LIST_PAGE_INDEX -> context.getString(R.string.plant_list_title)
-            else -> null
-        }
+        return tabTitles[position] ?: throw IndexOutOfBoundsException()
```
tiembo: OK to return null here, so just `tabTitles[position]` works

app/src/main/java/com/google/samples/apps/sunflower/HomeViewPagerFragment.kt
```diff
@@ -61,4 +61,16 @@ class ViewPagerFragment : Fragment() {
 
         return binding.root
     }
+
+    /** Return a hashmap from tab index to tab title to pass to SunflowerPagerAdapter **/
+    private fun getTabTitles(): HashMap<Int, String> {
```
XinyueZ: `SparseArrayCompat` would say

PR #515, opened 2019-09-22
Replace inproper '@+id' to '@id' in xml files

BlondeNamazu:
# Background

According to [official guide](https://developer.android.com/guide/topics/ui/declaring-layout?hl=JA#id), "@+id" should be used only in declarations. But this example uses "@+id" when referring other views. 

# Behaviour

This PR doesn't contain commits affect to app behaviour.


BlondeNamazu:
@googlebot I fixed it.

Issue #268, opened 2018-12-12
Spotless plugin not enforcing line lengths


tiembo:
Some code lines are greater than 100 characters, which is the suggested line length for Kotlin.

PR #517, opened 2019-09-29
#268: added rule to enforce max-lines-length

zhaonian:
- added rule to enforce max-lines-length
- fixed all the existing length issues in the project

Now, `./gradlew spotlessCheck` will fail if there is a line over 100 chars.

https://github.com/android/sunflower/issues/268


zhaonian:
thx for the review :D
Fixed wrong indentation and rebased/squashed all commits into 1.

tiembo:
Thank you!

PR #518, opened 2019-09-30
fix kotlin style issues

zhaonian:
What I did:
1. run `Analyze` -> `Inspect code` in Android Studio.
2. fix all the Kotlin style issues.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
@@ -43,11 +43,7 @@ class GardenPlantingAdapter :
     }
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
-        getItem(position).let { plantings ->
-            with(holder) {
-                bind(plantings)
-            }
-        }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/google/samples/apps/sunflower/adapters/GardenPlantingAdapter.kt
```diff
@@ -43,10 +43,8 @@ class GardenPlantingAdapter :
     }
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
-        getItem(position).let { plantings ->
-            with(holder) {
-                bind(plantings)
-            }
+        with(holder) {
+            bind(getItem(position))
```
nic0lette: This is good, but maybe we should also remove the `with` here, since there's only one line inside it. That makes this function really straight forward:

```
    override fun onBindViewHolder(holder: ViewHolder, position: Int) {
        holder.bind(getItem(position))
    }
```

zhaonian: great! thx!

zhaonian:
rebased/squash commits into 1.

nic0lette:
Thanks again!

PR #522, opened 2019-10-07
Clear Planted

nenio-traduko:
* Allows user to remove planted plants from their garden from the plant detail view.
* Edits pre-existing tests to match new logic.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
@@ -17,7 +17,6 @@
 package com.google.samples.apps.sunflower
 
 import android.content.Intent
-import android.graphics.drawable.Drawable
 import android.os.Build
 import android.os.Bundle
 import android.view.LayoutInflater
@@ -25,7 +24,6 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.app.ShareCompat
-import androidx.core.content.ContextCompat
 import androidx.core.widget.NestedScrollView
 import androidx.databinding.DataBindingUtil
 import androidx.fragment.app.Fragment
@@ -75,13 +73,6 @@ class PlantDetailFragment : Fragment() {
                         return R.string.removed_plant_from_garden
                     }
                     return R.string.added_plant_to_garden
-                }
-
-                override fun getToggleIcon(isPlanted: Boolean): Drawable? {
-                    if (isPlanted) {
-                        return ContextCompat.getDrawable(requireContext(), R.drawable.ic_clear)
-                    }
-                    return ContextCompat.getDrawable(requireContext(), R.drawable.ic_plus)
                 }
             }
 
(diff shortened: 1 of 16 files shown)


app/src/main/res/layout/fragment_plant_detail.xml
```diff
@@ -166,13 +166,12 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_margin="@dimen/fab_margin"
-            android:onClick="@{() -> callback.add(viewModel.plant)}"
+            android:onClick="@{() -> viewModel.isPlanted ?  callback.remove(viewModel.plant) : callback.add(viewModel.plant)}"
```
nic0lette: I wonder if it wouldn't be better to add this logic to the ViewModel itself, rather than having it here in the binding.

app/src/main/java/com/google/samples/apps/sunflower/PlantDetailFragment.kt
```diff
@@ -61,12 +61,21 @@ class PlantDetailFragment : Fragment() {
...
                     }
                 }
+
+                override fun remove(plant: Plant?) {
+                    plant?.let {
+                        plantDetailViewModel.removePlantFromGarden()
+                        Snackbar.make(
+                            root, R.string.removed_plant_from_garden, Snackbar.LENGTH_LONG
+                        )
+                            .show()
```
nic0lette: nit: Formatting

app/src/main/res/layout/fragment_plant_detail.xml
nenio-traduko: That would be better.

nenio-traduko:
@nic0lette It doesn't look like we are keeping a reference to the `GardenPlanting` object... I could instead create a transaction in the DAO that retrieves the `GardenPlanting` object and uses it to `delete`. Would that be better?

nenio-traduko:
@ashraf-patel I could also implement the toggle icon logic as a binding adapter. Let me know what your thoughts are on my approach.

ashraf-patel:
@nenio-traduko The logic is still not in the ViewModel, it is in the activity. I would suggest moving it to the ViewModel so that we can test it by exposing drawable resource ids from the ViewModel.

```
    val isPlanted = gardenPlantingRepository.isPlanted(plantId)
    val toggleIcon = Transformations.map(isPlanted) {
        if (it) R.drawable.ic_clear else R.drawable.ic_plus
    }
```

Your XML will look like this,
```
        <com.google.android.material.floatingactionbutton.FloatingActionButton
            android:id="@+id/fab"
            style="@style/Widget.MaterialComponents.FloatingActionButton"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:layout_margin="@dimen/fab_margin"
            android:onClick="@{() -> callback.toggle(viewModel.plant)}"
            --> android:src="@{viewModel.toggleIcon}" <-- No logic in UI
            android:tint="@android:color/white"
            app:layout_anchor="@id/appbar"
            app:layout_anchorGravity="bottom|end"
            app:shapeAppearance="@style/ShapeAppearance.Sunflower.FAB" />

```

You will need a binding adapter as well
```
@BindingAdapter("android:src")
fun setImageResource(imageView: ImageView, resource: Int) {
    imageView.setImageResource(resource);
}
```

ashraf-patel:
@nenio-traduko LGTM

PR #536, opened 2019-10-23
Update gradle version.

nic0lette:


app/src/main/res/values-v27/shape.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tiembo: Can you remove this file from this PR?

nic0lette: Done.

Issue #538, opened 2019-10-24
Filtering plant list causes strange visual list behavior


satur9nine:
Tested on mater with latest commit d1eda0b6a0687bbfb6d50b2499b73a5635ddeb5e

Switch to Plant List
Click filter button on top right
Click filter button on top right again
Scroll up and see first item is not top left then at top it re-animates and fixes itself in a strange way

![sunflower](https://user-images.githubusercontent.com/3181423/67530795-1ecf0100-f675-11e9-8255-16a2a5c43a90.png)

prasad79:
I have added a fix for this issue and made a pull request. 

however the test has failed with 
`Task :spotlessKotlin FAILED
Step 'ktlint' found problem in 'app/src/main/java/com/google/samples/apps/sunflower/PlantListFragment.kt':
Error on line: 53, column: 3
Exceeded max line length (100)
java.lang.AssertionError: Error on line: 53, column: 3`

PR #549, opened 2019-11-12
Fix for: Filtering plant list causes strange visual list behavior #538

prasad79:


arriolac:
Closing this as it is no longer relevant since the app was migrated to Compose

Issue #550, opened 2019-11-15
Crash when click two item in the list at the same time.


CodePoem:
As the title describes, the exception will happen when click two item  in the list at the same time.

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.google.samples.apps.sunflower, PID: 1795
    java.lang.IllegalArgumentException: navigation destination com.google.samples.apps.sunflower:id/action_view_pager_fragment_to_plant_detail_fragment is unknown to this NavController
        at androidx.navigation.NavController.navigate(NavController.java:804)
        at androidx.navigation.NavController.navigate(NavController.java:745)
        at androidx.navigation.NavController.navigate(NavController.java:731)
        at androidx.navigation.NavController.navigate(NavController.java:863)
        at com.google.samples.apps.sunflower.adapters.GardenPlantingAdapter$ViewHolder.navigateToPlant(GardenPlantingAdapter.kt:65)
        at com.google.samples.apps.sunflower.adapters.GardenPlantingAdapter$ViewHolder.access$navigateToPlant(GardenPlantingAdapter.kt:51)
        at com.google.samples.apps.sunflower.adapters.GardenPlantingAdapter$ViewHolder$1.onClick(GardenPlantingAdapter.kt:57)
        at android.view.View.performClick(View.java:6614)
        at android.view.View.performClickInternal(View.java:6591)
        at android.view.View.access$3100(View.java:786)
        at android.view.View$PerformClick.run(View.java:25948)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:201)
        at android.app.ActivityThread.main(ActivityThread.java:6806)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:547)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:873)
```

CodePoem:
I find a way to avoid this issue and create a pull request. #551

PR #551, opened 2019-11-15
Fix crash when click two item in the list at the same time. (#550)

CodePoem:
Fix issue #550


arriolac:
Closing this as it is no longer relevant since the UI was migrated to Compose

PR #556, opened 2019-11-17
Migrate Gradle files from groovy to the Gradle Kotlin DSL

hanleyt:
See slides showing benefits of this here: https://slides.com/tomashanley/migrating-to-gradl-kotlin-dsl/


.gitignore
```diff
@@ -4,7 +4,7 @@
 .idea/*
 !.idea/copyright
 .DS_Store
-/build
+build/
```
nic0lette: nit: If you want to do this, then potentially think about deleting `app/.gitignore` since all that file has in it is `/build`.

tiembo:
Hi @hanleyt , thanks for the PR!  While Android Studio support for Kotlin DSL script files is available starting in Studio 4.0 canary 1, we have a preference for making sure Sunflower works well with the latest Stable version of Studio (3.5.2 as of writing).

We can leave this PR open for now and revisit once the stable version of Studio adds support.  Feel free to also close this PR and reopen at a later time.

arriolac:
Done in #787

PR #565, opened 2019-12-18
Returns empty string instead of null

jacksondeng:


app/src/main/java/com/google/samples/apps/sunflower/HomeViewPagerFragment.kt
```diff
@@ -60,11 +60,11 @@ class HomeViewPagerFragment : Fragment() {
         }
     }
 
-    private fun getTabTitle(position: Int): String? {
+    private fun getTabTitle(position: Int): String {
         return when (position) {
             MY_GARDEN_PAGE_INDEX -> getString(R.string.my_garden_title)
             PLANT_LIST_PAGE_INDEX -> getString(R.string.plant_list_title)
-            else -> null
+            else -> ""
```
amardeshbd, line 67: Why though, whats the gain? 🙄

arriolac:
No longer valid since Sunflower was migrated to Compose

PR #568, opened 2019-12-29
Improve share message

jsnider:
Just a small addition I think makes the app better

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -32,7 +32,7 @@
     <string name="plant_date_header" translation_description="Text that is shown on each plant on the 'My Garden' screen which says the word Planted to show date of planting below">Planted</string>
     <string name="watered_date_header" translation_description="Text that is shown on each plant on the 'My Garden' screen which says the word Last Watered to show date of last watering below">Last Watered</string>
     <string name="menu_item_share_plant" translation_description="The label for the Share menu item">Share</string>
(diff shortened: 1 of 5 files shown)


jsnider:
@googlebot I signed it!

app/src/main/res/values/strings.xml
```diff
@@ -32,7 +32,7 @@
     <string name="plant_date_header" translation_description="Text that is shown on each plant on the 'My Garden' screen which says the word Planted to show date of planting below">Planted</string>
     <string name="watered_date_header" translation_description="Text that is shown on each plant on the 'My Garden' screen which says the word Last Watered to show date of last watering below">Last Watered</string>
     <string name="menu_item_share_plant" translation_description="The label for the Share menu item">Share</string>
-    <string name="share_text_plant" translation_description="The default messaged used when sharing a plant on SNS, etc.">Check out the %s plant in the Android Sunflower app</string>
+    <string name="share_text_plant" translation_description="The default messaged used when sharing a plant on SMS, etc.">Check out the %1$s plant in the Android Sunflower app \n%2$s</string>
```
piotrek1543: I think it should be in another Pull Request as it is not related to topic :-)

tiembo:
Thanks for the PR.  There are a additional things that are needed, such as migrating the database to include the new column and populating the column with the data (for both new and existing databases)

app/src/main/res/values/strings.xml
tiembo: SNS is short for social networking service

tiembo:
Closing due to inactivity.  Please comment if you'd like to resume discussion. Thanks!

PR #584, opened 2020-03-04
Use "let" language function instead of "val"

acious:
There is no need to use `val` variable.
I think that using `let` better than 'val' declaration


tiembo:
Thank you for the PR.  While using `.let` does leverage more Kotlin language features, separating out the `val` assignment and usage is a bit more clear in terms of readability and illustrating Navigation concepts.

PR #591, opened 2020-03-29
correct ktlint installation instructions

davidschof:


davidschof:
@googlebot I signed it!

README.md
```diff
@@ -112,7 +112,7 @@ from the ktlint [README](https://github.com/shyiko/ktlint/blob/master/README.md)
 
 - Inside the project root directory run:
 
-  `./ktlint --apply-to-idea-project --android`
+  `ktlint --apply-to-idea-project --android`
```
tiembo: Looks like these params are out of date; it's probably better to link directly to ktlin's README.  Let's change these lines to something like:

- Apply ktlint settings to Android Studio using these [instructions](https://github.com/pinterest/ktlint/blob/master/README.md#-with-intellij-idea)

davidschof: Thanks, I've made that change.

tiembo:
Thank you!

PR #608, opened 2020-05-13
Refactor

AbdoNasr3:
Some refactors, include:
-Replace ShareCompat with ShareSheet
-Update target, compile sdk versions
-move plant, GardenPlantings, PlantAndGardenPlantings to new package "models"


tiembo:
Thank you for the PR!  To make it easier to track changes, can you create two separate PRs, one for the SDK version and one for the ShareSheet?

Also, the Plant and other data classes are OK in their current locations.

PR #616, opened 2020-06-04
minor code fix

tedKim5178:
remove redundant local variable


tedKim5178:
@googlebot  I signed it!

PR #629, opened 2020-07-15
Add Paging 3.0 for gallery screen

tiembo:


lmorda:
Sorry to comment on an already merged PR, but I was wondering why `getSearchResultStream` no longer needs to be a suspending function?

PR #632, opened 2020-07-22
Replaced Glide by Coil

cbedoy:
Glide is written in Java, Coil is fully written with Kotlin, it uses coroutines and **Lightweight**


cbedoy:
@googlebot I signed it!

tiembo:
Hi there - thanks for the PR!  Right now [Glide is recommended](https://developer.android.com/topic/performance/graphics/load-bitmap) but Coil is definitely another popular image loading library.

PR #633, opened 2020-08-02
fix formatting so it passes ktlint (configured with -a option) at com…

ColoradoTim:
…mand line. Code formatting is a simple thing, but when someone new to the project gets a failed build on a clean copy of the repo it does not inspire confidence ;-)

background:  on MacOS, I installed ktlint (using brew), configured with -a (kotlin android style) option, run ./gradlew build, and build fails on several formatting problems. I changed the code so build succeeds, required changes (and nothing else) are in this PR


tiembo:
Thanks for the PR and catching this issue! I was able to reproduce the error when running `./gradlew build`.  We'll take a look at our toolchain / CI setup and make improvements to catch these issues.

PR #637, opened 2020-08-15
modify strings.xml

petprog:
I add some changes to strings.xml which may be implemented to other languages


tiembo:
Thanks for the PR. In general we want to only changes string resources when adding new content or making large changes for existing content (which should include translations).

PR #645, opened 2020-09-02
Remove kotlin stdlib dependency

skydoves:
The stdlib library – starting from Kotlin 1.4.0, the stdlib dependency is added automatically.

### [Dependency on the standard library added by default](https://kotlinlang.org/docs/reference/whatsnew14.html#dependency-on-the-standard-library-added-by-default)
You no longer need to declare a dependency on the stdlib library in any Kotlin Gradle project, including a multiplatform one. The dependency is added by default.

### Difference
- [x] Removed a kotlin stdlib dependency in the `app` module.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -82,7 +82,6 @@ dependencies {
     implementation "com.squareup.okhttp3:logging-interceptor:$rootProject.okhttpLoggingVersion"
     implementation "com.squareup.retrofit2:converter-gson:$rootProject.retrofitVersion"
     implementation "com.squareup.retrofit2:retrofit:$rootProject.retrofitVersion"
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$rootProject.kotlinVersion"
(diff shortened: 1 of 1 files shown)


nic0lette (changes requested):
Thanks. I think that this is pretty straight forward, but I think that if we are removing the `implementation` line we should also remove the variable definition from the top level [`build.gradle` file](https://github.com/android/sunflower/blob/87873648ec7cc24a6a4ddb540b786041cde09d22/build.gradle#L36).

skydoves:
Hi @nic0lette! Thanks for your PR review. 🙏
The `kotlinVersion` variable definition is also used in [kotlin-plugin](https://github.com/android/sunflower/blob/87873648ec7cc24a6a4ddb540b786041cde09d22/build.gradle#L62).

tiembo (approved):

nic0lette (approved):
You're right. Oops! Thank you!

tiembo:
Thanks @skydoves and @nic0lette !

PR #649, opened 2020-09-15
Added MotionLayout for transition animation

GeorgCantor:
[](https://postimg.cc/Cd4hzGpK)


arriolac:
Thanks for the contribution - Sunflower has been migrated to Compose so this PR no longer applies

PR #660, opened 2020-10-07
Implement plant list screen in compose

mr3y-the-programmer:
I've migrated partially plant list screen to compose, there is only a filter menu item that still described from XML

There is a little small caveat with this migration in comparison with the old approach:
-We've lost the animation of moving elements within recyclerView when clicking on filter icon to change the order of items


florina-muntenescu:
implemented meanwhile

PR #667, opened 2020-10-26
Update paging library for Kotlin extension

agustiyann:
Update a little bit, which might be missed.
That's about using the Kotlin extension for the Paging Library.


tiembo (approved):

tiembo:
Thanks for the PR!

PR #670, opened 2020-11-09
Updates Compose to alpha07

manuelvicnt:


app/build.gradle
```diff
@@ -88,7 +94,10 @@ dependencies {
     implementation "androidx.compose.ui:ui-viewbinding:$rootProject.composeVersion"
     implementation "androidx.ui:ui-tooling:$rootProject.composeVersion"
     implementation "androidx.compose.runtime:runtime-livedata:$rootProject.composeVersion"
-    implementation "com.google.android.material:compose-theme-adapter:$rootProject.composeVersion"
+//    implementation "com.google.android.material:compose-theme-adapter:$rootProject.composeVersion"
```
florina-muntenescu, line 97: should this be removed rather than commented out?

florina-muntenescu (approved):
Overall looks good. one extra question. feel free to merge after it's addressed

app/build.gradle
chrisbanes: I'd keep it, we'll need to revert back to this for the actual release.

chrisbanes (approved):

app/build.gradle
manuelvicnt: +1, I left it commented out so that we don't forget to revert it

tiembo (approved):

PR #671, opened 2020-11-12
Convert data layer to use Flow

tiembo:
Requires updating Room to 2.2.5.  Also updates Gradle version to 4.1.1


app/src/main/java/com/google/samples/apps/sunflower/data/PlantRepository.kt
```diff
@@ -16,6 +16,7 @@
 
 package com.google.samples.apps.sunflower.data
 
+import kotlinx.coroutines.flow.Flow
```
manuelvicnt, line 19: Is this import needed? I'd also add a comment about threading in this class.

e.g. `// Collecting from the flows in Dao is main-safe as Room's coroutines support moves the execution of the query off the main thread.`

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
```diff
@@ -35,26 +38,38 @@ class PlantListViewModel @ViewModelInject internal constructor(
     @Assisted private val savedStateHandle: SavedStateHandle
 ) : ViewModel() {
 
-    val plants: LiveData<List<Plant>> = getSavedGrowZoneNumber().switchMap {
-        if (it == NO_GROW_ZONE) {
-            plantRepository.getPlants()
-        } else {
-            plantRepository.getPlantsWithGrowZoneNumber(it)
-        }
+    private val growZoneChannel = ConflatedBroadcastChannel<Int>()
```
manuelvicnt, line 41: Interesting pattern that may arise here. I wasn't sure what this `growZoneChannel` was doing at the beginning but if my assumptions are correct, this is just basically an observable type that we want to get from `SavedStateHandle` and because it doesn't support Flows yet, you need to retrieve it using `savedStateHandle.getLiveData(GROW_ZONE_SAVED_STATE_KEY, NO_GROW_ZONE)`. Is that correct?

For that, we could use the new APIs introduced in coroutines 1.4.0 (`StateFlow` and `SharedFlow`). As you need to access the current value of this for the `isFiltered()` function, the API you're looking for is `StateFlow`.

```
private val growZone = getLiveData(GROW_ZONE_SAVED_STATE_KEY, NO_GROW_ZONE)
                            .asFlow() // convert the LiveData to Flow
                            .stateIn(viewModelScope) // convert the Flow to StateFlow
```

That should also react for when you set a new grow zone.

Then, `plants` gets simplified to `growZone.flatMapLatest { ... }.asLiveData()`.

build.gradle
```diff
@@ -46,7 +46,7 @@ buildscript {
         pagingVersion = '3.0.0-alpha02'
         recyclerViewVersion = '1.2.0-alpha04'
         retrofitVersion = '2.9.0'
-        roomVersion = '2.1.0'
+        roomVersion = '2.2.5'
```
manuelvicnt, line 49: Bump also coroutines to `1.4.0` :)

manuelvicnt (changes requested):

build.gradle
tiembo: Good catch!

app/src/main/java/com/google/samples/apps/sunflower/data/PlantRepository.kt
tiembo: Missed that unused import.  Good call on the comment - added in.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
tiembo: Yes, we'll need to convert the `LiveData` from `SavedStateHandle` to Flow.  Thanks for the pointers!

I couldn't determine how to correctly use `stateIn` so I used an approach with `MutableStateFlow`.  I think this is a step towards that direction, with the next step collapsing `getSavedGrowZoneNumberOrDefault` into `growZone` - is that the correct?

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
```diff
@@ -35,26 +42,37 @@ class PlantListViewModel @ViewModelInject internal constructor(
...
         savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, NO_GROW_ZONE)
     }
 
-    fun isFiltered() = getSavedGrowZoneNumber().value != NO_GROW_ZONE
+    fun isFiltered() = getSavedGrowZoneNumberOrDefault().value != NO_GROW_ZONE
 
-    private fun getSavedGrowZoneNumber(): MutableLiveData<Int> {
-        return savedStateHandle.getLiveData(GROW_ZONE_SAVED_STATE_KEY, NO_GROW_ZONE)
+    /** @return saved Grow Zone (if available); otherwise return [NO_GROW_ZONE] as the default */
+    private fun getSavedGrowZoneNumberOrDefault(): LiveData<Int> {
```
manuelvicnt, line 72: Seems like returning this LiveData is a waste as we're not assigning it to anything. Can we avoid this method and initialize `growZone` as follows?

```
private val growZone = MutableStateFlow(
  savedStateHandle.get(GROW_ZONE_SAVED_STATE_KEY) as Int ?: NO_GROW_ZONE
)
```

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
```diff
@@ -35,26 +42,37 @@ class PlantListViewModel @ViewModelInject internal constructor(
...
     }
 
     fun setGrowZoneNumber(num: Int) {
+        growZone.value = num
         savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, num)
     }
 
     fun clearGrowZoneNumber() {
+        growZone.value = NO_GROW_ZONE
         savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, NO_GROW_ZONE)
```
manuelvicnt, line 66: These `savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, zone)` could be simplified if we use the power of flow operators. What about using `.onEach` for that?

```
val plants: LiveData<List<Plant>> = growZone.flatMapLatest { zone ->
        // rest of the code
    }.onEach {
        savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, it)
    }
     .asLiveData()
```

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
```diff
@@ -35,26 +42,37 @@ class PlantListViewModel @ViewModelInject internal constructor(
...
     }
 
-    fun isFiltered() = getSavedGrowZoneNumber().value != NO_GROW_ZONE
+    fun isFiltered() = getSavedGrowZoneNumberOrDefault().value != NO_GROW_ZONE
```
manuelvicnt, line 69: This could use `growZone` instead.

```suggestion
    fun isFiltered() = growZone.value != NO_GROW_ZONE
```

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
tiembo: Done. This cleans up a lot!

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
tiembo: Done

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
tiembo: Conversation continued in subsequent thread.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
tiembo: Spoke offline - let's try moving the `.set` logic to the `init` block.

app/src/main/java/com/google/samples/apps/sunflower/viewmodels/PlantListViewModel.kt
```diff
@@ -35,27 +40,37 @@ class PlantListViewModel @ViewModelInject internal constructor(
...
         } else {
-            plantRepository.getPlantsWithGrowZoneNumber(it)
+            plantRepository.getPlantsWithGrowZoneNumber(zone)
+        }
+    }.asLiveData()
+
+    init {
+
+        // When `growZone` changes, store the new value in savedStateHandle
+        viewModelScope.launch {
```
manuelvicnt, line 58: As you're calling `collect`, you can consume the elements in the `collect`'s lambda block instead of using the `onEach` operator:

```
viewModelScope.launch {
    growZone.collect { newGrowZone ->
        savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, newGrowZone)
    }
}
```

Alternatively, you can avoid creating a new coroutine using the `launchIn` terminal operator. In that case, `onEach` is needed because `launchIn` doesn't take a lambda to consume the new element in the flow, it takes a `CoroutineScope` that uses to create a coroutine internally. 

```
growZone.onEach { newGrowZone ->
    savedStateHandle.set(GROW_ZONE_SAVED_STATE_KEY, newGrowZone)
}.launchIn(viewModelScope)
```

Both options are equivalent :) use the one you like most

tiembo: Chose the former approach and also added a description comparing the two based on the discussion here

build.gradle
```diff
@@ -27,11 +27,11 @@ buildscript {
         assistedInjectVersion = '0.5.2'
         constraintLayoutVersion = '2.0.0-beta3'
         coreTestingVersion = '2.0.0'
-        coroutinesVersion = "1.3.0-M2"
+        coroutinesVersion = "1.4.0"
```
manuelvicnt, line 30: Nit: This can be `1.4.2` already :)

manuelvicnt (approved):

build.gradle
tiembo: Done!

manuelvicnt:
Hi team! I think there are some tests that need to be updated.

Checked latest code in `main` and `androidTest`s assert for LiveData updates instead of Flow

tiembo:
Hi Manuel! Thanks for the heads up; I'll take a look at these

tiembo:
Tracking in #691

PR #679, opened 2020-12-21
chore: remove deprecated kotlin-android-extensions plugin

rosariopfernandes:
Not only is the kotlin-android-extensions deprecated, but it's also not being used in this project.


pfmaggi (approved):

tiembo (approved):

tiembo:
Thanks for the PR @rosariopfernandes !

PR #684, opened 2021-01-11
Replace "dataBinding = true" with "dataBinding true" in app/build.gradle

tatsuyafujisaki:
`=` is redundant here. It is slightly better to omit it for simplicity, as in the [official documentation](https://developer.android.com/topic/libraries/data-binding/start).


tiembo (approved):

tiembo:
Thank you for the PR!

PR #682, opened 2021-01-05
Removing blank files

Abhimanyu14:

PR #685, opened 2021-01-12
remove blank files

anmol-Git:
2 kotlin files are blank in the view model package


anmol-Git:
some files are blank

tiembo:
Hi there, I think #682 covers the same changes in this PR.  If that's the case, can you close this PR? Thank you!

PR #688, opened 2021-01-16
Improve Chinese support

Vixb1122:


Vixb1122:
@googlebot I signed it!

PR #689, opened 2021-01-16
Improve Chinese support

Vixb1122:


calren (approved):
Thanks for this @Vixb1122 !

PR #706, opened 2021-03-02
Move logic to ViewModel so it can be unit-tested

oliviadodge:
From a test-driven perspective it would be beneficial to get most of the logic covered in small unit tests since this is the fastest and most reliable way of testing an Android application. I believe a ViewModel can be unit-tested whereas a Fragment cannot, without heavy mocking.


AidaIssayeva (approved):

arriolac:
Already done in bea8e859

PR #726, opened 2021-06-30
Migrates PlantAdapter to Compose

manuelvicnt:


app/src/main/java/com/google/samples/apps/sunflower/adapters/ComposeRecyclerViewUtils.kt
```diff
@@ -0,0 +1,62 @@
...
+abstract class ComposeRecyclerViewAdapter<VH : ComposeViewHolder<*>> : RecyclerView.Adapter<VH>() {
+
+    override fun onViewRecycled(holder: VH) {
+        holder.composeView.disposeComposition()
+        super.onViewRecycled(holder)
+    }
+}
+
+abstract class ComposeViewHolder<T>(
+    val composeView: ComposeView
```
nickbutcher, line 45: Should we force extenders to construct the `ComposeView` or construct it internally, or even as a default? Might consumers want to configure the `ComposeView` they pass in?

manuelvicnt: We cannot construct it internally because `RecyclerView.ViewHolder` takes the `View` in its constructor. 

We could have it as a default but then it'd also require a `Context` to be passed in as a parameter. That might be even more confusing:

```
abstract class ComposeViewHolder<T>(
    context: Context,
    val composeView: ComposeView = ComposeView(context)
) : RecyclerView.ViewHolder(composeView) {
```

Not sure how often you want to configure the `ComposeView` passed in, maybe you want to change the layout params of it?

PR #732, opened 2021-07-23
Organize files into packages

AdilBolatkhanov:


arriolac:
No longer valid since migrating fully in Compose

PR #765, opened 2022-05-18
Update deprecated testing

wiryadev:
- replace junit `assertThat` with hamcrest one
- replace `ActivityTestRule` with `ActivityScenarioRule`
- replace `TestCoroutineDispatcher` with `TestDispatcher`

Note: if the test can not be run at all, downgrade **espresso** to 3.3.0 as i mentioned [here](https://github.com/android/sunflower/issues/764).


arriolac (approved):
Thanks!

PR #770, opened 2022-06-02
Removed redundant let call

sabiou:
Removed redundant call to let at

``binding.setClickListener {
                binding.plant?.let { plant ->
                    navigateToPlant(plant, it)
                }
            }``


arriolac:
Looks like this produced a build error @sabiou

PR #769, opened 2022-05-31
Merge compose branch into main

arriolac:
Merges `compose` branch into `main`.

## Changes
- Merge main into compose.
- Update Compose in RecyclerView implementation. (#766)
- [Compose]: Implement RecyclerView item in Compose. (#763)
- [Compose]: Update dependencies and remove reference to accompanist (#760)
- [Compose]: Upgrade to 1.2.0-alpha07 (#758)
- Update tooling dependencies (#754)
- [Compose] Update to rc02 (#731)
- [Compose] Update to beta07 (#716)
- Update to Compose beta02 (#708)
- @manuelvicnt Update to Compose beta01 (#704)
- Updates to Compose alpha12 (#697)
- Updates Compose to alpha11 (#693)
- Updates to Compose alpha10 (#683)
- Update to alpha08 version of Compose (#677)
- Updates Compose to alpha07 (#670)
- Updates Compose to alpha06 (#666)
- Updates to compose alpha05 (#664)
- Updates to Compose alpha04 (#657)
- Updates to Compose alpha (#639)
- Migrates Plant details screen to Compose (#1) (#638)

PR #772, opened 2022-06-09
Removed navigateToGallery function

AbdullahAleiti:
This function is seemingly redundant


arriolac:
This function is still needed but looks like the UI element that triggers navigating to the gallery was removed since #769 was integrated. 

I filed #773 to add that feature back.

PR #779, opened 2022-07-26
new animal tab

yanisapths:


app/src/main/java/com/google/samples/apps/sunflower/AnimalFragment.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val viewModel: AnimalViewModel by viewModels()
+
+    override fun onCreateView(
+            inflater: LayoutInflater,
+            container: ViewGroup?,
+            savedInstanceState: Bundle?
+    ): View {
+        val binding = FragmentAnimalBinding.inflate(inflater, container, false)
+
+        context ?: return binding.root
```
avinash--reddy, line 45: Could you please let me know why this null check is required ? In what cases the context can be null here ?

PR #781, opened 2022-08-01
Create SECURITY.md

andy13-pixel:


arriolac:
Looks like this file is just a template and not Sunflower-specific

PR #782, opened 2022-08-16
Correct PlantDetailViewModelTest testing by Espresso dependency

alidoran:
The below error is shown on the PlantDetailViewModelTest running.

`FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:checkDebugAndroidTestDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class org.checkerframework.checker.compilermsgs.qual.CompilerMessageKey found in modules checker-3.1 
     
This is fixed by Gradle changing.
Also, remove a duplicate data binding and buildFeatures.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,9 +23,6 @@ apply plugin: 'androidx.navigation.safeargs.kotlin'
 
 android {
     compileSdkVersion rootProject.compileSdkVersion
-    buildFeatures {
-        dataBinding true
-    }
     defaultConfig {
         applicationId "com.google.samples.apps.sunflower"
         minSdkVersion rootProject.minSdkVersion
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -23,9 +23,6 @@ apply plugin: 'androidx.navigation.safeargs.kotlin'
 
 android {
     compileSdkVersion rootProject.compileSdkVersion
-    buildFeatures {
-        dataBinding true
-    }
```
arriolac, line 28: Revert. Some layouts still use data binding.

arriolac (changes requested):

app/build.gradle
alidoran: Is it need to have two buildFeatures in Gradle and two dataBinding true?

arriolac: Oh, gotcha. I missed your comment that this is a duplicate.

arriolac:
Please merge `main` again into your fork

arriolac:
Looks like this is no longer an issue.

PR #794, opened 2022-09-19
Fix %d appearing in watering needs string.

### LINKED ISSUE #778: String Format Specifier visible at runtime
Did I do something wrong? ("%d" should be replaced by a value at runtime):
every-percent-d-days

sitamshrijal:
Fixes #778


arriolac:
Thanks!

arriolac (approved):

PR #798, opened 2022-10-10
Replace Coil with Glide.

arriolac:


app/src/main/java/com/google/samples/apps/sunflower/compose/plantdetail/PlantDetailView.kt
```diff
@@ -271,35 +285,56 @@ private fun PlantDetailsContent(
...
-    if (painter.state is AsyncImagePainter.State.Loading) {
-        Box(
+            .height(imageHeight)) {
+        if (isLoading) {
+            Box(
+                Modifier
+                    .fillMaxSize()
+                    .background(placeholderColor)
+            )
+        }
```
bentrengrove, line 307: Glide doesn't have some sort of placeholder functionality built in?

bentrengrove (approved):

app/src/main/java/com/google/samples/apps/sunflower/compose/plantdetail/PlantDetailView.kt
arriolac: Ah - that just got added recently so will update this PR

arriolac: Looks like it's not released yet. Will add a comment to update it when that feature is available on maven.

bentrengrove: Sounds good!

PR #817, opened 2022-12-20
Add MigrationJourney.md

arriolac:


docs/MigrationJourney.md
```diff
@@ -0,0 +1,42 @@
...
+
+As of 2022, Sunflower demonstrates a partially migrated View-based app. The intent is to fully migrate Sunflower so that the UI layer is only in Compose.
+
+This document captures the high-level strategy, as well as links to issues/pull requests for migration steps, taken to migrate the rest of the app to be Compose-only.
+
+The general steps followed to migrate Sunflower to Compose are:
+
+1. Planning the migration approach
+2. Migrate existing screens to Compose one by one
+3. Migrate Navigation Component to Compose
```
bentrengrove, line 13: I think you need to explicitly mention that you are replacing fragments in step 3 and using ComposeView inside a fragment until then. Just a sentence somewhere in this doc. Maybe just change step 3 title to "Remove fragment usage and replace Jetpack Navigation with Navigation Compose"

bentrengrove (approved):
Just one suggestion but lgtm

docs/MigrationJourney.md
arriolac: Good call! Added a note on that

PR #825, opened 2023-01-12
Call ReportDrawn instead of ReportDrawnWhen { true }

mlykotom:
This landed yesterday as part of alpha03


arriolac (approved):
Thanks @mlykotom!

PR #847, opened 2023-05-09
Update home_screen.xml

buddywhitman:
Positioned the toolbar below the status bar while maintaining the window's translucent status.
[Source 1](https://stackoverflow.com/questions/28807874/android-toolbar-status-bar-overlapping#:~:text=as%20Virus%20correctly%20says%2C%20the%20answer%20is%20here%3A,xml%20item%20and%20it%20will%20work%20fine%20%3A%29), [Source 2](https://stackoverflow.com/questions/29738510/toolbar-overlapping-below-status-bar)


arriolac:
Closed by #858

Issue #842, opened 2023-04-23
Top app bar is overlapping with the status bar


s12dar:
The top app bar overlaps with the status bar, making it unable to click the filter action button.

Steps to reproduce the issue:
->  Click on any plant item in the garden or plant list.
-> Navigate back.

<table>
  <tr>
    <td>
      <img src="https://user-images.githubusercontent.com/39422653/233814160-212d679a-9f0c-4160-aa7a-f849e437d23e.png" alt="Screenshot_20230422-201810" width="200"><br>
      &emsp;&emsp;&emsp;&emsp;&emsp;Then
    </td>
    <td>
      <img src="https://user-images.githubusercontent.com/39422653/233814403-2ecae5e1-8390-45a5-b9c9-b6365ad11439.png" alt="Screenshot_20230422-201810" width="200"><br>
      &emsp;&emsp;&emsp;&emsp;&emsp;After
    </td>
  </tr>
</table>

Device: Pixel 6
Android version: UpsideDownCake

Ps. Would be even better if we make the status bar transparent. 
(Issue can be assigned to me)

arriolac:
Thanks for reporting! Feel free to send us a PR

s12dar:
Sure, go ahead.

s12dar:
> Can I do it? This is going to be my first open source contribution.

KwakEuiJin:
If this issue is not solved, can I solve it?

arriolac:
Yes, feel free to send us a PR

KwakEuiJin:
Thanks😊

I solved it through two methods.

1. Migration of top app bar to compose

2. I set the padding value according to the size of status bar.

But the first way isn't to use scroll animation from appbar

I will PR each of the two code🙂

arriolac:
Looked into this and I believe this might be framework bug. Filed here: https://issuetracker.google.com/issues/287315739

PR #839, opened 2023-03-29
Meghan/m3

MagicalMeghan:
Pull request for migrating Sunflower to M3

PR #848, opened 2023-05-09
Fix Issue 842: #842 AppBar overlap error resolve

KwakEuiJin:
issue: #842 

I fixed 4 bugs through Compose migration.

1.  overlaps statusbar and TopAppbar
2. Appbar title A bug that is not fixed in the center when this menu is added
3. does not maintain the tab of the previous screen when clicking the back button
4. filter button not action

However, the code did not apply Material 3, so it was difficult to add an animation in which the appbar disappeared at the bottom scroll. 

https://github.com/android/sunflower/assets/93872496/88b93e5b-3165-4a68-a7d9-eeecd4cd2f44


app/src/main/java/com/google/samples/apps/sunflower/GardenActivity.kt
```diff
@@ -62,20 +35,7 @@ class GardenActivity : AppCompatActivity() {
 
         setContent {
             MdcTheme {
-                SunflowerApp(
```
arriolac, line 65: This implementation should not have been changed—the reason interop is used in this manner is so that the toolbar can collapse when you scroll in the plant list screen. With this change, the toolbar no longer collapses. I describe in this blog post the issue: https://medium.com/androiddevelopers/migrating-sunflower-to-jetpack-compose-f840fa3b9985

arriolac (changes requested):
Collapsing toolbar no longer working. The change here should be reverted—your other PR is closer to what we want.

Also FYI that Sunflower is actively being migrated to M3 (see #839). The `m3` branch will eventually land on `main` once the migration is complete. That PR should fix #842

app/src/main/java/com/google/samples/apps/sunflower/GardenActivity.kt
KwakEuiJin: Yes then please check #843번 PR #842 We will close the PR.

PR #849, opened 2023-05-10
Fix issue #842 second way

KwakEuiJin:
issue : #842 
In this PR, I solved the bug that overlaps statusbar and appbar by settinging the padding value.

This way, scroll animation at the bottom of the app bar works, but other bugs could not be fixed.

You can see the PR that fixed other bugs in this link. https://github.com/android/sunflower/pull/848

However, the PR of this link has the disadvantage of not being able to operate the animation of the app bar. 

https://github.com/android/sunflower/assets/93872496/1c2d8abf-fcc6-4cea-8535-1793a8509de4

### DIFF
--- a/app/src/main/res/layout/home_screen.xml
+++ b/app/src/main/res/layout/home_screen.xml
@@ -25,7 +25,7 @@
       android:id="@+id/coordinator_layout"
       android:layout_width="match_parent"
       android:layout_height="match_parent"
-      android:fitsSystemWindows="true">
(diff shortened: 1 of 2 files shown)


arriolac (changes requested):
https://developer.android.com/develop/ui/views/layout/edge-to-edge#handle-overlaps should be followed for handling insets

KwakEuiJin:
This is the best way to correct the error. I don't know how to modify it more here.
I think using material3 is the best way.

PR #859, opened 2023-06-15
Handling insets in Compose and not in Views.

arriolac:
Similar effect to #858 but in this change, I'm adding insets support on the Compose side instead of the Views side.
This follows our guidance where the outermost layout (which in Sunflower is a `ComposeView`) should consume the insets.

Context: https://developer.android.com/jetpack/compose/layouts/insets#interop-view


arriolac:
While this is technically the ideal fix, this presents new problems due to the way `CoordinatorLayout` also controls system bars. Namely, it requires that `fitsSystemsWindow` be set to `true` so that the system bar scrim can be applied when the toolbar collapses. And for that to work alongside the proper insets, the top-level ComposeView needs to set `consumeWindowInsets` to false so that `CoordinatorLayout` can consume it instead.

Issue #840, opened 2023-04-21
App Crash when scroll down in Gallery Screen


HieuNguyennd96:
When I scroll down in Gallery screen, app is crashed immediately 
Device: Pixel 2XL 
Android Version 11
Log: 
FATAL EXCEPTION: main
                                                                                                    Process: com.google.samples.apps.sunflower, PID: 17655
                                                                                                    java.lang.IllegalArgumentException: Key 8Jau7nWrgmE was already used. If you are using LazyColumn/Row please make sure you provide a unique key for each item.
                                                                                                    	at androidx.compose.ui.layout.LayoutNodeSubcompositionsState.subcompose(SubcomposeLayout.kt:411)
                                                                                                    	at androidx.compose.ui.layout.LayoutNodeSubcompositionsState$Scope.subcompose(SubcomposeLayout.kt:733)
                                                                                                    	at androidx.compose.foundation.lazy.layout.LazyLayoutMeasureScopeImpl.measure-0kLqBqw(LazyLayoutMeasureScope.kt:118)
                                                                                                    	at androidx.compose.foundation.lazy.grid.LazyMeasuredItemProvider.getAndMeasure-ednRnyU(LazyMeasuredItemProvider.kt:44)
                                                                                                    	at androidx.compose.foundation.lazy.grid.LazyMeasuredLineProvider.getAndMeasure-bKFJvoY(LazyMeasuredLineProvider.kt:70)
                                                                                                    	at androidx.compose.foundation.lazy.grid.LazyGridMeasureKt.measureLazyGrid-0cYbdkg(LazyGridMeasure.kt:147)
                                                                                                    	at androidx.compose.foundation.lazy.grid.LazyGridKt$rememberLazyGridMeasurePolicy$1$1.invoke-0kLqBqw(LazyGrid.kt:329)
                                                                                                    	at androidx.compose.foundation.lazy.grid.LazyGridKt$rememberLazyGridMeasurePolicy$1$1.invoke(LazyGrid.kt:179)
                                                                                                    	at androidx.compose.foundation.lazy.layout.LazyLayoutKt$LazyLayout$1$2$1.invoke-0kLqBqw(LazyLayout.kt:71)
                                                                                                    	at androidx.compose.foundation.lazy.layout.LazyLayoutKt$LazyLayout$1$2$1.invoke(LazyLayout.kt:69)
[...]

PR #875, opened 2023-06-22
Fix: crash when scroll in Gallery Screen

qamarelsafadi:
### [#840](https://github.com/android/sunflower/issues/840) Issue :
App Crashes when start scrolling down using LazyVerticalGrid

### ✅ Solution
In Compose when we use key inside items we should make sure that this key is a unique key, so I append the index with the photo.id to make sure that it's unique.

Till next PR 🚀!


arriolac (approved):
Thanks for this!

leinardi:
Hi, I ended up in this PR because I'm facing similar issues (the back-end I'm using is returning duplicate items that, obviously, share the same ID) but I'm not sure about this fix: isn't including the index inside the key kind of invalidating the entire purpose of setting a **stable** and unique key? From what I understood, the key of an item should not change if the same item is simply moving position (changing the index) inside the list but, if you add the index as part of the key, this won't be true. Isn't this solution the same as not setting the key at all, if in the end the key that you are using is not stable?

PR #877, opened 2023-06-22
 Change Bar Color

mohamed-abd-elnabi:
Part of GTC open source intiative


bentrengrove:
Hello

This isn't something we would want on our home page README so I am going to close this, sorry. 

Please in the future file an issue and add details on why you want to make a change and we can confirm if it is something we want to proceed with.

PR #879, opened 2023-06-22
Fix wrong licensing information.

yenerm:
We use Apache license but we do not license software to apache foundation. Also no license info is needed in the readme file


arriolac (approved):
LGTM

Issue #853, opened 2023-05-18
App crashes.


hhyeok1026:
**How to reproduce:**
Add a plant → Close the app. (Clean up the app by pressing the recent button) → Run the app again (crash occurs)

- Both the emulator and the real device crashed.

- But, after deleting the cache data of the app, it works again when relaunched.

**error stack**
2023-05-18 12:07:50.140 13418-13418 AndroidRuntime          com.google.samples.apps.sunflower    E  FATAL EXCEPTION: main
Process: com.google.samples.apps.sunflower, PID: 13418
java.lang.IllegalStateException: removeReporter() called when all reporters have already been removed.
at androidx.activity.FullyDrawnReporter.removeReporter(FullyDrawnReporter.kt:107)
at androidx.activity.compose.ReportDrawnComposition.removeReporter(ReportDrawn.kt:74)
at androidx.activity.compose.ReportDrawnKt$ReportDrawnWhen$1$invoke$$inlined$onDispose$2.dispose(Effects.kt:484)
at androidx.compose.runtime.DisposableEffectImpl.onForgotten(Effects.kt:85)
at androidx.compose.runtime.CompositionImpl$RememberEventDispatcher.dispatchRememberObservers(Composition.kt:1094)
at androidx.compose.runtime.CompositionImpl.applyChangesInLocked(Composition.kt:820)
at androidx.compose.runtime.CompositionImpl.applyChanges(Composition.kt:842)
at androidx.compose.runtime.Recomposer$runRecomposeAndApplyChanges$2$2.invoke(Recomposer.kt:592)
at androidx.compose.runtime.Recomposer$runRecomposeAndApplyChanges$2$2.invoke(Recomposer.kt:510)
at androidx.compose.ui.platform.AndroidUiFrameClock$withFrameNanos$2$callback$1.doFrame(AndroidUiFrameClock.android.kt:34)
at androidx.compose.ui.platform.AndroidUiDispatcher.performFrameDispatch(AndroidUiDispatcher.android.kt:109)
at androidx.compose.ui.platform.AndroidUiDispatcher.access$performFrameDispatch(AndroidUiDispatcher.android.kt:41)
at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.doFrame(AndroidUiDispatcher.android.kt:69)
at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1229)
at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1239)
[...]

PR #880, opened 2023-06-23
[Fix] removeReporter App Crashes 

qamarelsafadi:
## What?
I've updated the compose activity dependency version to 1.7.2
#853

## Why?
This change prevents the removeReporter() called when all reporters have already been removed exception which causes the app crashes. See #853 for more information.

## How?
The 1.7.2 version prevents this crash by fixing the bug which is caused by onDismiss and observeReporter().

## 

This PR is part of [GTC](https://www.facebook.com/groups/gazatechcommunity/) open source initiative

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,7 +16,7 @@
 [versions]
 accessibilityTestFramework = "4.0.0"
 accompanist = "0.28.0"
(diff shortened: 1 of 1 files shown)


arriolac (approved):
Thanks! This closes #853

Issue #898, opened 2023-07-04
[Bug]: Incorrect packaging.yaml


Keworker:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is there a StackOverflow question about this issue?

- [X] I have searched StackOverflow

### Is this an issue related to the sample app?

- [X] Yes, this is a specific issue related to this samples repo.

### What happened?

When I searched for examples for Jetpack Data Store in [documentation](https://developer.android.com/topic/libraries/architecture/datastore), I found this repository. This repo has this tag in [sunflower/.google/packaging.yaml](https://github.com/android/sunflower/blob/main/.google/packaging.yaml):
`  - JetpackDataStore`
However, I cant find usage of Data Store in this app. I think, that this tag should be deleted.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Keworker:
https://github.com/android/sunflower/pull/907
I created a Pull Request to solve the problem.

Keworker:
After merging PR it is steel here https://developer.android.com/topic/libraries/architecture/datastore
I think, that it need some time to be updated.

PR #907, opened 2023-07-29
Fix Incorrect packaging.yaml #898 

Keworker:
This project doesn't use Jetpack DataStore, but while it had the JetpackDataStore tag it was incorrectly recommended as a hint in the DataStore documentation. 

I remove tag Jetpack DataStore.


riggaroo (approved):
thanks!

PR #910, opened 2023-08-02
Replace Insert with Upsert at PlantDao.

Keworker:
PlantDao use Insert with provided on conflict strategy. With newer version of Room it's better to use Upsert in this case.


arriolac (approved):
LGTM. Thanks!

PR #911, opened 2023-08-06
Replace additional function with custom matcher for better assertion errors.

Keworker:
Before this PR if `GardenPlantingTest` throws an error, it were possible only to see errors like this:
```
java.lang.AssertionError: 
Expected: <2024>
     but: was <2023>
	 
java.lang.AssertionError: 
Expected: <9>
     but: was <8>
	 
java.lang.AssertionError: 
Expected: <7>
     but: was <6>
```
I add custom matcher and now it's easier to read assertion errors stack traces.
```
java.lang.AssertionError: 
Expected: 06.08.2024
     but: was 06.08.2023
	 
java.lang.AssertionError: 
Expected: 06.09.2023
     but: was 06.08.2023
	 
java.lang.AssertionError: 
Expected: 07.08.2023
     but: was 06.08.2023
```
Additional advantage of this way is that asserts become more readable. 

I also add visibility modifier _internal_ for all unit-test classes, because they should have less visibility.


Keworker:
@mlykotom , excuse me, are you here?

Keworker:
I'm sorry, but i think that @mlykotom won't review it. @arriolac , can you review this PR, please?

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
```diff
@@ -0,0 +1,66 @@
...
+         * Creates a matcher for [Calendar]s that only matches when year, month and day of
+         * actual calendar are equal to year, month and day of expected calendar.
+         *
+         * For example:
+         * <code>assertThat(someDate, hasSameDateWith(Calendar.getInstance()))</code>
+         *
+         * @param expected calendar that has expected year, month and day [Calendar]
+         */
+        @Factory
+        fun hasSameDateWith(expected: Calendar): Matcher<Calendar> = HasSameDateWith(expected)
```
arriolac, line 64: Consider renaming to `equalTo`

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val expected: Calendar
+) : TypeSafeDiagnosingMatcher<Calendar>() {
+    private val formatter = SimpleDateFormat("dd.MM.yyyy")
+
+    override fun describeTo(description: Description?) {
+        description?.appendText(formatter.format(expected.time))
+    }
+
+    override fun matchesSafely(actual: Calendar?, mismatchDescription: Description?): Boolean {
+        return (actual?.let {
```
arriolac, line 43: This is a bit difficult to read. I recommend returning early using an `if` statement to check if `actual` is null

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
```diff
@@ -0,0 +1,66 @@
...
+            actual.get(YEAR) == expected.get(YEAR) &&
+                    actual.get(MONTH) == expected.get(MONTH) &&
+                    actual.get(DAY_OF_MONTH) == expected.get(DAY_OF_MONTH)
+        } ?: false).also { _ ->
+            mismatchDescription?.appendText("was ")
```
arriolac, line 48: This will always append text even when `actual` matches `expected` - it should only build `mismatchDescription` if the two don't match.

Keworker: I changed it, now we add it only if we need. But why shouldn't we do it in normal situation ? As far as I know it doesn't affect anything if the test is successful.

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
Keworker: It was fixed

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
Keworker: Renamed. Maybe it would better set name like `equalDateTo` ? Equal to sounds really abstract, it can conflict with over matchers in feature.

Keworker:
@arriolac excuse me, can you check this PR again (resolve conversations if problems are solved or left new comments) ?

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
arriolac: I was thinking more along the lines of:

```
if (actual == null)
  return false


if (actual.get(YEAR) == expected.get(YEAR) && actual.get(MONTH) == expected.get(MONTH) && actual.get(DAY_OF_MONTH) == expected.get(DAY_OF_MONTH)) 
  return true


mismatchDescription?.appendText("was ")?.appendText(actual?.time?.let { formatter.format(it) } ?: "null")
return false
}
```

app/src/test/java/com/google/samples/apps/sunflower/test/HasSameDateWith.kt
arriolac: It doesn't affect the validity of the test, but it does affect what's printed out in the console

arriolac (changes requested):
Please make AGP version update in a separate PR. Otherwise, LGTM

gradle/libs.versions.toml
```diff
@@ -16,7 +16,7 @@
 [versions]
 accessibilityTestFramework = "4.0.0"
 activityCompose = "1.7.2"
-androidGradlePlugin = "8.0.2"
+androidGradlePlugin = "8.1.1"
```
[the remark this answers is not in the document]
Keworker: Resolved: 
https://github.com/android/sunflower/pull/923

arriolac (approved):
LGTM

Issue #917, opened 2023-09-15
[Update]: Migrate [Hilt & Room] from KAPT to KSP


anselmoalexandre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Describe the problem

Keeping in mind that this is a Kotlin project, using KSP in favour of KAPT, makes much easier to understand Kotlin - specific features, makes much easier to develop lightweight compiler plugins and also it can run twice faster than KAPT annotation processor, etc.

### Describe the solution

Setup the migration from KAPT to KSP 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Code-indomitus:
Hi can I work on this?

arriolac:
Please feel free to send us a PR :)

PR #921, opened 2023-10-13
Issue-917: Introduced the ksp plugin in project dagger and hilt depen…

Code-indomitus:
…dencies now use ksp instead of kapt. Removed kapt plugin and use overall in project. Changed kotlin version from '1.8.0' to '1.9.10'. Changed hilt version from 2.44.2 to 2.48 as it is the version that supports ksp. Had to change compose-compiler version to 1.5.3 as a result of upgrading kotlin version to '1.9.10'.

Issue-917: Introduced the ksp plugin in project dagger and hilt dependencies now use ksp instead of kapt. Removed kapt plugin and use overall in project. Changed kotlin version from '1.8.0' to '1.9.10'. Changed hilt version from 2.44.2 to 2.48 as it is the version that supports ksp. Had to change compose-compiler version to 1.5.3 as a result of upgrading kotlin version to '1.9.10'.


arriolac (approved):
Thanks for this!

arriolac:
Fixes #917

Issue #918, opened 2023-09-15
[Update]: Clean up Version Catalogs and upgrade dependencies flavours 


anselmoalexandre:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is this a feature request for the samples?

- [X] Yes, this is a specific request related to this samples repo.

### Describe the problem

The version catalog is not well organised and does not help when it comes to adding new plugins and dependencies since it's not clean and not properly organised.

Also, the dependencies versions are outdated which make it difficult to use more recent and lightweights improvements on some apis.

### Describe the solution

- Organise Version catalogs
- Introduce bundles and plugins sections
- Upgrade dependency versions to latest flavours

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Hau2202:
Tuyệt vời

PR #925, opened 2023-10-28
- add plugins section to version catalog, update gradle build files (#918)

borednaeem:


mlykotom (approved):
Thanks for this change!

Issue #863, opened 2023-06-19
Separate navigation route and args from SunflowerApp to specific file


qamarelsafadi:

arriolac:
Please add more details in the description. What are the pros/cons of this change?

qamarelsafadi:
Done, 

Let me know what you think about it.

PR #931, opened 2023-11-16
Separate navigation routes and args issue #863

sahilsk3333:
Separate navigation routes and args issue #863

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/utilities/Screen.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/utilities/Screen.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/google/samples/apps/sunflower/utilities/Screen.kt
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package com.google.samples.apps.sunflower.utilities
+
+import androidx.navigation.NamedNavArgument
+import androidx.navigation.NavType
+import androidx.navigation.navArgument
+
+sealed class Screen(
```
arriolac, line 23: Rather than putting this under the `utilities` package, I recommend adding this in `compose` package.

arriolac (commented):
Thanks! Just one small suggestion on the file's location. Other than that, LGTM.

For reference for other folks reading this PR, you can also reference now NiA structures routes/screens. For example: https://github.com/android/nowinandroid/blob/main/feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksScreen.kt#L80

app/src/main/java/com/google/samples/apps/sunflower/utilities/Screen.kt
sahilsk3333: Thanks, I've addressed the feedback in commit 'Moved Screen class to compose package'. Could you take a look, @arriolac ?

[Commit Link](https://github.com/sahilsk3333/sunflower/commit/e6a9ea1eaa4a23d1fa76eb7d4179a09670dfce7b)

arriolac (approved):

PR #933, opened 2023-11-21
Enable e2e.

arriolac:
Enable edge-to-edge on Sunflower.

### DIFF
--- a/app/src/main/java/com/google/samples/apps/sunflower/compose/home/HomeScreen.kt
+++ b/app/src/main/java/com/google/samples/apps/sunflower/compose/home/HomeScreen.kt
@@ -16,6 +16,7 @@
 
 package com.google.samples.apps.sunflower.compose.home
 
+import android.util.Log
 import androidx.annotation.DrawableRes
 import androidx.annotation.StringRes
 import androidx.compose.foundation.ExperimentalFoundationApi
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/google/samples/apps/sunflower/compose/home/HomeScreen.kt
```diff
@@ -85,11 +85,11 @@ fun HomeScreen(
                 scrollBehavior = scrollBehavior
             )
         }
-    ) {
+    ) { contentPadding ->
         HomePagerScreen(
             onPlantClick = onPlantClick,
             pagerState = pagerState,
-            modifier = Modifier.padding(it)
+            Modifier.padding(top = contentPadding.calculateTopPadding())
```
riggaroo, line 92: Did you add padding to the bottom of the screen too so it can scroll the content out from underneath the bar?

arriolac: There's already padding applied so it offsets correctly when you scroll all the way down in the plant list screen: https://github.com/android/sunflower/blob/main/app/src/main/java/com/google/samples/apps/sunflower/compose/plantlist/PlantListScreen.kt#L58

...though technically it doesn't take into account `navigationBarsPadding` so let me add that.

riggaroo (approved):

PR #655, opened 2020-09-23
Added Portuguese string translation - value-pt

anselmoalexandre:
Hi @tiembo 
I hope you okay.

I was watching your video explaining how to use **Jetpack library Components** using **Sunflower Android app**, and I decided to help adding **_Portuguese language_** to this app which I found it helpful to control our plants needs.

Please, review and possibly merge this pull request into the master.
Thank you.

PR #936, opened 2023-12-13
[Task] Update dependencies and clean version catalog

anselmoalexandre:
Hi team,

In this PR, I have cleanup and versions catalog file, introducing [bundles] block packaging together group of dependencies of the same flavour offering me dependency readability.

Please review
Thanks


bentrengrove:
Thanks for the contribution, this looks like multiple unrelated changes grouped together which means we wouldn't be able to merge it.

- There are what look to be test changes in the HomeScreen file
- I don't think we need the gradle file reformatting
- The Portuguese strings file

We would love adding the Portuguese strings to the app, could you please create a new PR just with those changes? The rest I don't think we could accept unfortunately as I don't see the benefit.

anselmoalexandre:
> Thanks for the contribution, this looks like multiple unrelated changes grouped together which means we wouldn't be able to merge it.
> 
> * There are what look to be test changes in the HomeScreen file
> * I don't think we need the gradle file reformatting
> * The Portuguese strings file
> 
> We would love adding the Portuguese strings to the app, could you please create a new PR just with those changes? The rest I don't think we could accept unfortunately as I don't see the benefit.

Thanks @bentrengrove for the feedback.
About the multiple unrelated changes, my bad on that, I actually branched from my old PR which is #655 for the Portuguese translation I did a while back. Kindly take a look on it.

Now, reformatting the version catalog file to introducing bundles, make the gradle dependencies more readable and combining related group libraries together. 

Thanks

PR #937, opened 2023-12-14
 plantImage not expanding 

yelaaaaaaaaa:
Avoid the plantImage not expanding all the way when you swipe up and down quickly.


arriolac (approved):
Thanks for the fix! Please sign the CLA and I'll go ahead and merge it

PR #939, opened 2023-12-15
feat:Avoid the plantImage not expanding all the way when you swipe up…

yelaaaaaaaaa:
Avoid the plantImage not expanding all the way when you swipe up and down quickly.


arriolac (approved):
LGTM

PR #941, opened 2023-12-21
Update Glide implementation and item key ID.

arriolac:
* Update Glide to the latest version. The latest version also removed support for passing another composable as a placeholder for performance reasons. So I've updated the placeholder to show a color instead.
* Fixed a crash in the gallery screen - regression introduced when updating to paging compose


arriolac:
Fixes #942

MagicalMeghan (approved):

PR #945, opened 2024-01-02
Remove deprecated SunflowerImage Glide wrapper in favor of GlideImage

sabiou:
The SunflowerImage() wrapper was introduced as a workaround for the issue https://github.com/bumptech/glide/issues/4977, and now that the issue has been resolved, there is no need for this wrapper in the codebase.

Other minor changes:

- Simplifying the expression `if (plant != null && isPlanted != null && showSnackbar != null)` to `if (plant != null && showSnackbar != null)` as `(isPlanted != null)` is always true.

- Replacing deprecated Icons.Filled.ArrowBack with Icons.AutoMirrored.Filled.ArrowBack is a good practice to address the deprecation and use the updated alternative.


arriolac (approved):
Thanks! Looks like previews were also broken. I'd like to merge in #946 (which fixes previews) before integrating this

sabiou:
> Thanks! Looks like previews were also broken. I'd like to merge in #946 (which fixes previews) before integrating this

Good. Thank you @arriolac

arriolac (approved):
LGTM

PR #948, opened 2024-01-03
Update dependencies and add update deps workflow

arriolac:
Changes:
* Update dependencies to latest and update breaking changes
* Add a new workflow that runs every month to check for updated dependencies (it can also be manually run)

### DIFF
--- a/.github/ci-gradle.properties
+++ b/.github/ci-gradle.properties
@@ -1,0 +1,26 @@
+#
+# Copyright 2020 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+org.gradle.daemon=false
+org.gradle.parallel=true
+org.gradle.jvmargs=-Xmx5120m
+org.gradle.workers.max=2
+
+kotlin.incremental=false
+kotlin.compiler.execution.strategy=in-process
+
+# Controls KotlinOptions.allWarningsAsErrors. This is used in CI and can be set in local properties.
+warningsAsErrors=true
--- a/.github/ci-gradle.properties
+++ b/.github/ci-gradle.properties
@@ -1,5 +1,5 @@
 #
-# Copyright 2020 The Android Open Source Project
+# Copyright 2024 The Android Open Source Project
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/buildscripts/init.gradle.kts
+++ b/buildscripts/init.gradle.kts
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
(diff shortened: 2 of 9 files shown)


.github/ci-gradle.properties
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright 2020 The Android Open Source Project
```
IanGClifton, line 2: I mean, it still feels like 2020 but supposedly it's 2024 these days

buildscripts/init.gradle.kts
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
IanGClifton, line 2: 2022 is closer haha

buildscripts/toml-updater-config.gradle
```diff
@@ -26,3 +26,21 @@ versionCatalogUpdate {
         keepUnusedPlugins.set(true)
     }
 }
+
+def isNonStable = { String version ->
+    def stableKeyword = ['RELEASE', 'FINAL', 'GA'].any { it -> version.toUpperCase().contains(it) }
```
IanGClifton, line 31: Where do these (potentially) show up? Is it reading something from maven? Otherwise, I'd think none of our versions would match since the stable releases are things like 1.2.3.

gradle/libs.versions.toml
```diff
@@ -14,55 +14,58 @@
...
-material3 = "1.2.0-alpha11"
+material3 = "1.2.0-beta01"
 # @keep
-minSdk = "23"
-monitor = "1.6.0"
-navigation = "2.5.3"
-okhttpLogging = "4.10.0"
-pagingCompose = "1.0.0-alpha19"
-profileInstaller = "1.2.0"
+minSdk = "28"
```
IanGClifton, line 46: This is a pretty big jump; intentional?

gradle/libs.versions.toml
```diff
@@ -93,11 +96,11 @@ androidx-paging-compose = { module = "androidx.paging:paging-compose", version.r
 androidx-profileinstaller = { module = "androidx.profileinstaller:profileinstaller", version.ref = "profileInstaller" }
 androidx-room-compiler = { module = "androidx.room:room-compiler", version.ref = "room" }
 androidx-room-ktx = { module = "androidx.room:room-ktx", version.ref = "room" }
-androidx-room-runtime = { module = "androidx.room:room-runtime", version.ref = "room" }
+androidx-room-runtime = "androidx.room:room-runtime:2.5.2"
```
IanGClifton, line 99: Why does this have a specific version instead of the room value above (2.6.1 now)?

buildscripts/toml-updater-config.gradle
arriolac: Oh good call—we don't need this line. This was a copy from https://github.com/android/compose-samples/blob/main/JetNews/buildscripts/toml-updater-config.gradle#L30

gradle/libs.versions.toml
arriolac: I think I bumped this while debugging, but it doesn't have to be 28. Will revert.

gradle/libs.versions.toml
arriolac: The `version-catalog-update-plugin` decided to update this and it's unclear why 🤔  However, this dependency is not used by the project so we can remove it altogether.

IanGClifton (approved):
LGTM, but I'll leave it up to you on whether you want another set of eyes on it

arriolac:
Thanks! Will merge this one in as I'd like to update dependencies since that fixes some issues such as previews not working (regression introduced in an alpha version of material3). I'll create a separate PR if anyone else on the team finds anything

PR #951, opened 2024-01-24
🤖 Update Dependencies

arriolac:


gradle/libs.versions.toml
```diff
@@ -16,34 +16,34 @@
...
+androidGradlePlugin = "8.2.2"
+benchmark = "1.2.3"
 # @keep
 compileSdk = "34"
 # @keep
 compose-compiler = "1.5.7"
-composeBom = "2023.10.01"
-composeLatest = "1.6.0-beta03"
+composeBom = "2024.01.00"
+composeLatest = "1.7.0-alpha01"
```
bentrengrove, line 26: Do you actually want this?

arriolac: Some dependencies in Sunflower use `composeLatest` though tbh I'm not sure why it couldn't use stable instead. Let me see if I can get those to use stable instead

arriolac: Done in 866ed7a

bentrengrove (approved):

PR #953, opened 2024-02-08
comment update

Lucodivo:
ViewModel is used in PlantDetailsScreen not PlantDetailFragment. Probably missed during migration from Fragments to Compose.


arriolac (approved):
Thanks!

